PR #9, opened 2018-07-16
Implement shared object repositories

mdekstrand:


mdekstrand:
I am having difficulty getting the Plasma support for this to actually work, so it looks unlikely to be useful.

PR #48, opened 2018-11-29
Add Random recommender

KimuraTian:
Add a random recommender in `basic` module. The random recommender randomly recommends items to a user from all items or candidate items (if provided). 
To control the randomness, an integer seed or a pandas Series indexed by user ids can be passed to initialize the random recommender.


lenskit/algorithms/basic.py
```diff
@@ -314,3 +314,41 @@ def load_model(self, path):
...
+
+    def __init__(self, random_state=None):
+        """
+        Args:
+             random_state: int or Series, optional
+                Seed/Seeds for the random sampling of items. If int, then recommending random items
+                for each user with the same seed. If Series, then it is the seeds for the users,
+                indexed by user id.
+        """
+        self.random_state = random_state
```
mdekstrand: Does the random state serve any purpose other than testing?

tests/test_basic.py
```diff
@@ -201,3 +201,86 @@ def test_pop_save_load(tmp_path):
     assert recs.score.iloc[0] == counts.max()
     # the 10 most popular should be the same
     assert all(counts.index[:10] == recs.item[:10])
+
+
+def test_random():
```
mdekstrand: Would it be easy enough to split this in to 2-3 test functions so it isn't just one big long one?

mdekstrand:
Superseded by #125.

PR #49, opened 2018-11-29
Optimize ALS

mdekstrand:
This uses proper solvers for ALS, and takes better advantage of sparsity in ALS-I.

### DIFF
--- a/lenskit/algorithms/als.py
+++ b/lenskit/algorithms/als.py
@@ -13,7 +13,7 @@ from .. import util
 _logger = logging.getLogger(__name__)
 
 
-@njit(parallel=True, nogil=True)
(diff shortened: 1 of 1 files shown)


mdekstrand:
@benfred would you mind giving this a look to see if I'm doing something silly?

benfred:
if you need sample code for the dposv code, implicit uses it here https://github.com/benfred/implicit/blob/master/implicit/_als.pyx#L112 for doing the same calculation.

mdekstrand:
Thanks! I like this `dposv` idea - it also seems it will reduce matrix/vector copies & allocation. Not a bottleneck but always nice to reduce.

PR #58, opened 2019-01-15
Update getting started for python script

MartijnMillecamp:
I copied the getting started into a python script and added the changes

### DIFF
--- a/doc/ALS.py
+++ b/doc/ALS.py
@@ -1,0 +1,98 @@
+import pandas as pd
+import numpy as np
+from lenskit import batch, topn, util
+from lenskit import crossfold as xf
+from lenskit.algorithms import als
+from lenskit.algorithms import Recommender
+
+# read in the movielens 100k ratings with pandas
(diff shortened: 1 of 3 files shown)


mdekstrand:
Thank you for your contribution!

The purpose of this PR is unclear to me. It adds some comments, which are helpful, but it looks like its primary functional outcome is to make tables display in their textual form instead of using Jupyter's table display - why might that be desirable?

MartijnMillecamp:
If you want to get started and you copy paste the code in a normal python script, it is not working :(
Probably it is more useful to provide this script to download instead of changing the getting started?

mdekstrand:
Thank you again for this, and I apologize for the radio silence. Given the changes in LensKit and in our Getting Started scripts, I think it is probably not worth the time to bring this up to merge-ready status.

Non-Jupyter runs of Lenskit are being added to https://github.com/lenskit/lk-demo-experiment.

PR #70, opened 2019-02-19
Drop deprecated CondaEnvironment task

mdekstrand:
The CondaEnvironment task is now deprecated (see Microsoft/azure-pipelines-tasks#9573); this uses the [Yaml script method](https://docs.microsoft.com/en-us/azure/devops/pipelines/languages/anaconda?view=azure-devops).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -306,7 +306,7 @@ jobs:
 
   - script: |
       conda create -n lkpy -qy $(conda.deps)
-      conda install -n lkpy llvm-openmp
+      conda install -n lkpy -qy llvm-openmp
     displayName: Set up Conda environment
 
   - script: |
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -290,15 +301,12 @@ jobs:
...
-  - task: CondaEnvironment@1
-    inputs:
-      createCustomEnvironment: true
-      environmentName: lkpy
-      packageSpecs: $(conda.deps)
-      updateConda: false
+  - bash: echo "##vso[task.prependpath]$CONDA/bin"
+    displayName: Add conda to PATH
+
+  - script: conda update -y conda
```
brcrista: I think you need to `sudo` this call because of where conda is installed on Hosted macOS

mdekstrand: Thanks! I'm going to just drop this update line; we'll live with a possibly outdated conda unless that starts causing problems.

PR #72, opened 2019-02-25
fixed fallback __str__() + test

keener101:
Added a test for checking str(algos) with fallback, and fixed issue.

### DIFF
--- a/lenskit/algorithms/basic.py
+++ b/lenskit/algorithms/basic.py
@@ -263,6 +263,7 @@ class Fallback(Predictor):
         str_algos = [str(algo) for algo in self.algorithms]
         return 'Fallback([{}])'.format(', '.join(str_algos))
 
+
 class TopN(Recommender):
     """
     Basic recommender that implements top-N recommendation using a predictor.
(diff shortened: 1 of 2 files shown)


mdekstrand:
I am very confused by the test status.

tests/test_basic.py
```diff
@@ -96,6 +96,9 @@ def test_fallback_list():
     assert isinstance(params['algorithms'][0], basic.Memorized)
     assert isinstance(params['algorithms'][1], basic.Bias)
 
+def test_fallback_string():
+    algo = basic.Fallback([basic.Memorized(simple_df), basic.Bias()])
+    str(algo)
```
mdekstrand: This test doesn't have any assertions - two ways I like to fix that:

- if the purpose is just to make sure that it doesn't fail, then add a comment after this line, e.g. `# str succesful, that's what we need`
- we can make the test a little more advanced - e.g. test that the string begins with `Fallback`.

lenskit/algorithms/basic.py
```diff
@@ -260,7 +260,9 @@ def predict_for_user(self, user, items, ratings=None):
         return preds.reindex(items)
 
     def __str__(self):
-        return 'Fallback([{}])'.format(', '.join(self.algorithms))
+
```
mdekstrand: If we're revising this, drop unnecessary blank line here (I don't think it adds clarity).

Issue #83, opened 2019-03-07
Rewrite docs for recommend=None and n=None for topn


keener101:
Docs imply that `n=None` (`batch.recommend`) and `recommend=None` (`multieval`) will result in complete recommendation lists. This is not the case. Will try with `n=0` / `n=-1` and see if that works.

keener101:
n = -1 results in the correct functionality.

mdekstrand:
So, I like `recommend=None` for that. What if we used `recommend=False` to turn off recommendation? Is it too confusing to have a false-y value (e.g. `None`) enable recommendation with unlimited lists?

keener101:
I think so. The idea from the meeting (I think) was..

All false-y values (None, False, 0) = turn off recommendation
Number values (1, 2, 100) = that size of recommendation list
True = full recommendation list

mdekstrand:
@keener101 Yes, that is correct!

PR #87, opened 2019-03-09
first stab at fixing recs size errors

keener101:
I expect to get some substantial 'fix this' feedback on this push.

The fix itself seems simple enough - just expand the checks to turn off recommendations to also include `0` and `False` (in addition to the previously checked `None`), and then after that, set `recommend=None` if `True` is passed in. 

I feel a bit out of my depth currently with the test classes when they get this complex. I copied the pattern from other tests that utilized `MultiEval`. 

`test_sweep_norecs` creates a `MultiEval` with `recommend = None` and makes sure no `recommendations.parquet` file is created. It then *repeats* the process with `recommend = False`. I wasn't sure if you'd want this or if it's overkill (especially given how slow these tests are). I could have also added one for `recommend = 0`, but didn't.

There's a bunch of additional `assert` lines that aren't really needed to test this - but I figured I'd leave them in during the first submission in case you'd prefer to check them regardless. 

`test_sweep_allrecs` simply makes sure a `recommendations.parquet` file is created when `recommend=True` is passed. **However it does not check if full recommendations are generated**. I wasn't sure how I would check this - I imagine it would work with an ``assert len(recs) == X`` check, but I'm not sure exactly what that size would be. (I have some confusion about ``ml_pandas.renamed.ratings`` and where exactly it is coming from).

### DIFF
--- a/lenskit/batch/_multi.py
+++ b/lenskit/batch/_multi.py
@@ -291,7 +291,7 @@ class MultiEval:
         return preds, watch.elapsed()
 
     def _recommend(self, rid, algo, test, candidates):
-        if not self.recommend: #if recommend is any false-y val (0, None, False), turn off recs
+        if not self.recommend:  #if recommend is any false-y val (0, None, False), turn off recs
             return None, None
         elif self.recommend is True:  # special value True means unlimited
             nrecs = None
--- a/lenskit/batch/_multi.py
+++ b/lenskit/batch/_multi.py
@@ -291,16 +291,17 @@ class MultiEval:
         return preds, watch.elapsed()
 
     def _recommend(self, rid, algo, test, candidates):
-        if self.recommend is None or self.recommend is False or self.recommend is 0:
+        if not self.recommend: #if recommend is any false-y val (0, None, False), turn off recs
             return None, None
-
-        if self.recommend is True:
-            self.recommend = None
+        elif self.recommend is True:  # special value True means unlimited
+            nrecs = None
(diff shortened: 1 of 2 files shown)


lenskit/batch/_multi.py
```diff
@@ -291,9 +291,12 @@ def _predict(self, rid, algo, test):
         return preds, watch.elapsed()
 
     def _recommend(self, rid, algo, test, candidates):
-        if self.recommend is None:
+        if self.recommend is None or self.recommend is False or self.recommend is 0:
```
mdekstrand: why not just `if not self.recommend`? That covers all the falsey values.

lenskit/batch/_multi.py
```diff
@@ -291,9 +291,12 @@ def _predict(self, rid, algo, test):
...
             return None, None
 
+        if self.recommend is True:
+            self.recommend = None
```
mdekstrand: This is going to change the recommend list size for the next run-through, not what we want. We can instead do a local variable:

```python
if not self.recommend:
    return stuff
elif self.recommend is True:  # special value True means unlimited
    nrecs = None
else:  # recommend has rec size
    nrecs = self.recommend
```

lenskit/batch/_multi.py
keener101: I... did not realize that. How elegant.

mdekstrand:
To link things: this addresses #83.

Issue #86, opened 2019-03-09
user-user aggregate 'sum' doesn't actually sum


mdekstrand:
Right now, setting `aggregate` in UserKNN doesn't actually _do_ anything.

PR #88, opened 2019-03-12
Actually use weighted sum (#86)

mdekstrand:


mdekstrand:
The Python35 build failure is due to a temporary dependency issue (numba/llvmlite#471).

PR #89, opened 2019-03-16
added documentation

keener101:
simple doc change

### DIFF
--- a/lenskit/batch/_multi.py
+++ b/lenskit/batch/_multi.py
@@ -41,8 +41,9 @@ class MultiEval:
         predict(bool):
             whether to generate rating predictions.
         recommend(int):
-            the number of recommendations to generate per user. Any false-y value (None,
-            False, 0) will disable top-n. True will generate the full recommendation list.
+            the number of recommendations to generate per user. Any false-y value (``None``,
+            ``False``, ``0``) will disable top-n. The literal value ``True`` will generate recommendation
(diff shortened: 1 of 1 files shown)


lenskit/batch/_multi.py
```diff
@@ -41,14 +41,15 @@ class MultiEval:
         predict(bool):
             whether to generate rating predictions.
         recommend(int):
-            the number of recommendations to generate per user (None to disable top-N).
+            the number of recommendations to generate per user. Any false-y value (None,
+            False, 0) will disable top-n. True will generate the full recommendation list.
```
mdekstrand: Let's use a little syntax here - `None` and `False` are literal code values, so they should be formatted as such; the syntax for that is

    ``None``

Documentation uses reStructuredText, not Markdown, so it is a little different.  Same for `True`, and let's specifically say "The literal value `True` will generate recommendation lists of unlimited size."

mdekstrand:
Little formatting.

If you want to preview docs locally, you can do that with (this sequence _should_ install all doc deps):

    conda install invoke sphinx sphinx_rtd_theme notebook ipython
    pip install recommonmark nbsphinx
    invoke docs

This will put docs under `build/docs`, and you can open them in your browser.

keener101:
I get a `Can't find any collection named 'tasks'!` output when I try to `invoke docs`.

Issue #21, opened 2018-07-31
Add shims for Surprise


mdekstrand:

sherif-med:
any ideas for this issue, what are the proposed approaches ?

mdekstrand:
What needs to happen here is to implement the [LensKit algorithm interfaces](http://lkpy.lenskit.org/en/stable/interfaces.html) on top of Surprise, translating the data into the form Surprise expects and translating the results into LensKit's result formats.

The Implicit and HPF modules have examples of other shims. We just need to do a comparable thing for Surprise, and to add tests that exercise the new code.

PR #109, opened 2019-06-27
Surprise shims

sherif-med:
Adding surprise shims with tests

### DIFF
--- a/lenskit/algorithms/surprise.py
+++ b/lenskit/algorithms/surprise.py
@@ -1,0 +1,100 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Tue Jun 25 12:09:10 2019
+
+@author: mohamed.cherif
+"""
+
+import logging
+import pandas as pd
+import numpy as np
+import inspect
+
+from ..matrix import ratings_surprise
+from . import Recommender
+
+_logger = logging.getLogger(__name__)
+
+
+
+class SurpriseBaseRec(Recommender):
+    """
+    Base class for surprise-backed recommenders.
+    Args:
+        delegate(implicit.RecommenderBase):
+            The delegate algorithm.
+    Attributes:
+        delegate(implicit.RecommenderBase):
+            The :py:mod:`surprise` delegate algorithm.
+        matrix_(scipy.sparse.csr_matrix):
+            The user-item rating matrix.
+        user_index_(pandas.Index):
+            The user index.
+        item_index_(pandas.Index):
+            The item index.
+    """
+
+    def __init__(self, delegate):
+        self.delegate = delegate
+
+    def fit(self, ratings):
+        dataset = ratings_surprise(ratings)
+        trainset = dataset.build_full_trainset()
+
+        _logger.info('training %s ', self.delegate)
+
+        self.delegate.fit(trainset)
+        
+        self.ratings = ratings
+        self.trainset = trainset
+        self.inner_user_index_ = trainset.all_users()
+        self.inner_item_index_ = trainset.all_items()
+
+        return self
+
+    def recommend(self, user, n=None, ratings=None):
+        
+        ##Getting user inner id
+        inner_uid = self.trainset.to_inner_uid(user)
+        # getting all items
+        all_items = self.ratings["item"].unique()
+        # getting already rated items
+        rated_items = self.ratings.loc[self.ratings["user"] == user, "item"]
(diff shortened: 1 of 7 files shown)


mdekstrand:
Thanks for submitting this! From a very quick look, it looks like you definitely have the right code structure & this is on a good path. I'm currently on vacation, will plan to look at it more closely when I return to work next week.

mdekstrand:
For cross-linking: this PR works on #21.

lenskit/algorithms/surprise.py
```diff
@@ -0,0 +1,235 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+
+class SurpriseBaseRec(Recommender):
+    """
+    Base class for surprise-backed recommenders.
+    Args:
+        delegate(implicit.RecommenderBase):
```
mdekstrand: The docstring should be updated to reference Surprise in the types.

lenskit/algorithms/surprise.py
```diff
@@ -0,0 +1,235 @@
...
+        trainset = dataset.build_full_trainset()
+
+        _logger.info('training %s ', self.delegate)
+
+        self.delegate.fit(trainset)
+        
+        self.ratings = ratings
+        self.trainset = trainset
+        self.inner_user_index_ = trainset.all_users()
+        self.inner_item_index_ = trainset.all_items()
```
mdekstrand: This all makes sense.

lenskit/algorithms/surprise.py
```diff
@@ -0,0 +1,235 @@
...
+
+        return self
+
+    def recommend(self, user, n=None, ratings=None):
```
mdekstrand: Given this implementation, it looks like this class should implement `Predictor` instead of `Recommender`, so that it can be combined with other candidate selectors. The logic in this method is a repeat of what `UnratedItemsCandidateSelector` does.

lenskit/matrix.py
```diff
@@ -490,3 +491,20 @@ def sparse_ratings(ratings, scipy=False):
...
+    
+    Args:
+        ratings(pandas.DataFrame): a data table of (user, item, rating) triples.
+        
+    Returns:
+        dataset:
+            a surprise dataset instance
+    """
+    reader = Reader()
+    dataset = Dataset.load_from_df(ratings.loc[:, ['user', 'item', "rating"]], reader)
```
mdekstrand: Any reason for not being consistent in string syntax?

setup.py
```diff
@@ -50,6 +50,9 @@
         ],
         'implicit': [
             'implicit'
+        ],
+        'surprise': [
+            'surprise'
```
mdekstrand: Good!

tests/test_surprise.py
```diff
@@ -0,0 +1,180 @@
...
+surprise_algo_list = [SurpriseKNNBasic, SurpriseCoClustering,
+ SurpriseKNNBaseline, SurpriseKNNWithMeans, SurpriseKNNWithZScore, SurpriseNMF,
+ SurpriseNormalPredictor, SurpriseSlopeOne, SurpriseSVD, SurpriseSVDpp]
+# none knn based algos and none normal predictor list for test
+surprise_algo_list_none_knn_based = [SurpriseCoClustering, SurpriseNMF
+                                     , SurpriseSlopeOne, SurpriseSVD, SurpriseSVDpp]
+
+@mark.slow
+@mark.skipif(not have_surprise, reason='surprise not installed')
+def test_surprise_train_rec():
```
mdekstrand: Since this is operating over each of the algorithms, it looks like a good candidate for `mark.parametrize` to parameterize it over the algorithms. That way each algorithm can fail independently.

tests/test_surprise.py
```diff
@@ -0,0 +1,180 @@
...
+        r2 = a2.recommend(100, n=20)
+        assert len(r2) == 20
+        assert all(r2 == recs)
+
+
+@mark.slow
+@mark.eval
+@mark.skipif(not have_surprise, reason='surprise not installed')
+@mark.skipif(not lktu.ml100k.available, reason='ML100K data not present')
+def test_surprise_batch_accuracy():
```
mdekstrand: Same here.

mdekstrand:
I would love to get this merged, but we have quite a bit of updating to do on it.

PR #118, opened 2019-08-20
Fix Getting Started link

teej:


hsm207:
Is this fix live?

https://lkpy.lenskit.org/en/stable/  still links to https://lkpy.lenskit.org/en/stable/quickstart.html

mdekstrand:
It is live in https://lkpy.lenskit.org/en/latest/. Will cut a new release in the next couple of weeks.

PR #173, opened 2020-06-19
Feature/114 als run time train

### LINKED ISSUE #114: Add run-time feature training for ALS
The ALS algorithms are readily amenable to computing new or updated feature vectors for new users on the fly: just solve the least squares problem for the user's feature vector given their ratings and the item feature matrix.

We just need to write the code, and possibly factor out part of the solver to allow single-vector solutions. This is probably most robust with the LU-decomposition solver, although we could run the coordinate descent solver for a few rounds.

Unlike #60, this pertains to what we do at predict or recommend time, not updating the model itself.

carlos10seg:
Added _train_bias_row_lu function and updated predict_for_user function to allow new ratings.

Four new unit tests for ALS:

- test_als_predict_basic_for_new_ratings
- test_als_predict_basic_for_new_user_with_new_ratings
- test_als_predict_for_new_user_with_new_ratings (this maybe could be removed)
- test_als_predict_for_new_users_with_new_ratings

Closes #114


lenskit/algorithms/als.py
```diff
@@ -126,6 +126,32 @@ def _train_matrix_lu(mat: _CSR, this: np.ndarray, other: np.ndarray, reg: float)
...
+def _train_bias_row_lu(items, ratings, other, reg):
+    """
+    Args:
+        items(np.ndarray[i64]): the item IDs the user has rated
+        ratings(np.ndarray): the user's (normalized) ratings for those items
+        other(np.ndarray): the item-feature matrix
+        reg(float): the regularization term
+
+    Returns:
+        np.ndarray: the user-feature vector (equivalent to V in the current LU code)
```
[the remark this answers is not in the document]
mdekstrand: Are you using Visual Studio Code? If so, install the EditorConfig extension so it can read the `.editorconfig` file and prevent these kinds of issues in the future.

lenskit/algorithms/als.py
```diff
@@ -466,8 +492,16 @@ def _train_iters(self, current, uctx, ictx):
...
 
     def predict_for_user(self, user, items, ratings=None):
-        # look up user index
-        return self.score_by_ids(user, items)
+        if ratings is not None:
+            # get the indexes for the items in ratings
+            indexes = []
+            for i in ratings.index: # ratings has as index the item id
+                indexes.append(self.item_index_.get_loc(i))
+            user_feature = _train_bias_row_lu(indexes, ratings.values, self.item_features_, self.regularization)
```
mdekstrand, line 500: Here is the problem. Rows are trained on _normalized_ ratings - you need to subtract the bias out of the ratings before training the row.

lenskit/algorithms/mf_common.py
```diff
@@ -67,7 +67,7 @@ def lookup_items(self, items):
         """
         return self.item_index_.get_indexer(items)
 
-    def score(self, user, items):
+    def score(self, user, items, user_feature=None):
```
mdekstrand, line 70: This is a good design for this. `user_feature` is a pretty long name, though.

lenskit/algorithms/als.py
mdekstrand: You can also make this more efficient, and need to deal with an issue where a rated item isn't in the item index.

Quick way to do this:

```python
ri_idxes = self.item_index_.get_indexer_for(ratings.index.values)
ri_good = ri_idxes >= 0
ri_items = ri_idxes[ri_good]
ri_values = ratings.values[ri_good]
```

Then you can pass `ri_items` and `ri_values` to `_train_bias_row_lu`.

lenskit/algorithms/mf_common.py
```diff
@@ -141,12 +143,12 @@ def score(self, user, items, raw=False):
...
         """
 
-        rv = super().score(user, items)
+        rv = super().score(user, items, user_feature)
 
         if not raw:
             # add bias back in
             rv = rv + self.global_bias_
-            if self.user_bias_ is not None:
+            if self.user_bias_ is not None and user is not None:
```
mdekstrand, line 151: So: for a new user, this means it won't add _this_ user's bias back in.

The clean solution to this problem is to do a bit of updating on the `Bias` class, and use it instead of reproducing bias-oriented logic in the matrix factorizers. Issue #174 discusses such changes.

tests/test_als_explicit.py
```diff
@@ -82,6 +82,80 @@ def test_als_predict_basic(m):
     assert preds.loc[3] >= -0.1
     assert preds.loc[3] <= 5.1
 
+@methods
```
mdekstrand, line 85: I don't think we need to parameterize each of these over `@methods`. We can just use the default solver to train the initial model.

tests/test_als_explicit.py
```diff
@@ -82,6 +82,80 @@ def test_als_predict_basic(m):
...
+def test_als_predict_basic_for_new_ratings(m):
+    algo = als.BiasedMF(20, iterations=10, method=m)
+    algo.fit(simple_df)
+
+    assert algo.global_bias_ == approx(simple_df.rating.mean())
+
+    new_ratings = pd.DataFrame({'item': [1, 2],
+                            'rating': [4.0, 5.0]}).set_index('item')
```
mdekstrand, line 93: I would make this a series:

```python
new_ratings = pd.Series([4.0, 5.0], index=[1, 2])
```

tests/test_als_explicit.py
```diff
@@ -82,6 +82,80 @@ def test_als_predict_basic(m):
...
+
+    new_u_id = -1
+    new_ratings = u_ratings.set_index('item').drop(columns=['user'])   
+    print(new_ratings) 
+    new_preds = algo.predict_for_user(new_u_id, [i], new_ratings)
+
+    assert preds.loc[i] == approx(new_preds.loc[i])
+
+@methods
+def test_als_predict_for_new_user_with_new_ratings(m):
```
mdekstrand, line 122: What is the difference between this and the previous test?

tests/test_als_explicit.py
```diff
@@ -82,6 +82,80 @@ def test_als_predict_basic(m):
...
+    items = np.random.choice(ratings.item.unique(), 100)
+    new_u_id = -1
+
+    algo = als.BiasedMF(20, iterations=10, method=m)
+    algo.fit(ratings)
+    
+    for u in users:
+        preds = algo.predict_for_user(u, items)  
+
+        new_ratings = ratings[ratings.user == u].set_index('item').drop(columns=['user', 'timestamp'])
```
mdekstrand, line 154: make a series:

```python
new_ratings = ratings[ratings.user == u].set_index('item')['rating']
```

tests/test_als_explicit.py
```diff
@@ -82,6 +82,80 @@ def test_als_predict_basic(m):
...
+        new_preds = algo.predict_for_user(new_u_id, items, new_ratings)
+
+        for i in items:
+            assert preds.loc[i] == approx(new_preds.loc[i])
```
mdekstrand, line 158: You can test the whole thing at one go:

```python
assert new_preds.values == approx(preds.values)
```

mdekstrand:
Merging this PR will close #114.

lenskit/algorithms/als.py
```diff
@@ -126,6 +126,30 @@ def _train_matrix_lu(mat, this: np.ndarray, other: np.ndarray, reg: float):
     return np.sqrt(frob)
 
 
+@njit(parallel=True, nogil=True)
```
mdekstrand, line 129: This doesn't need `parallel=True`.

lenskit/algorithms/als.py
```diff
@@ -412,19 +439,20 @@ def _normalize(self, ratings, users, items):
```
mdekstrand, line 439: Do we still need lines 429-439? Do we still need to return the bias tuple?

lenskit/algorithms/als.py
```diff
@@ -465,8 +493,26 @@ def _train_iters(self, current, uctx, ictx):
...
+            u_feat = _train_bias_row_lu(ri_it, ri_val, self.item_features_, self.regularization)
+            scores = self.score_by_ids(user, items, u_feat)
+
+            if self.bias:
+                return self.bias.inverse_transform_user(user, scores, u_offset)
+            else:
+                return scores
+        else:
+            # look up user index
+            return self.score_by_ids(user, items)
```
mdekstrand, line 515: Ok, so we aren't being consistent in de-biasing here. If `score_by_ids` de-biases, then when we have new ratings we debias twice; if it does not, then when we don't have new ratings we don't debias.

lenskit/algorithms/als.py
```diff
@@ -465,14 +466,30 @@ def _train_iters(self, current, uctx, ictx):
...
+            if self.bias:
+                ratings, u_offset = self.bias.transform_user(ratings)
+
+            ri_idxes = self.item_index_.get_indexer_for(ratings.index)
+            ri_good = ri_idxes >= 0
+            ri_it = ri_idxes[ri_good]
+            ri_val = ratings.values[ri_good]
+
+            u_feat = _train_bias_row_lu(ri_it, ri_val, self.item_features_, self.regularization)
+            print("user_features from predic_for_user: " + str(u_feat))
```
mdekstrand, line 481: Here we have another stray `print`. Either remote of change to `_log.debug`. I think maybe remove?

PR #181, opened 2020-07-07
Add transform_user and inverse_transform_user methods to bias.

### LINKED ISSUE #174: Add per-user transform APIs to bias
The Bias model would benefit from per-user transform APIs, in addition to its bulk transform API.

These methods should be:

```python
def transform_user(self, ratings: pd.Series) -> (pd.Series, float):
    """
    Transform a user's ratings by subtracting the bias model.  Returns the transformed ratings and the user bias.
    """

def inverse_transform_user(self, user, ratings: pd.Series, user_bias: float or None = None) -> pd.Series:
    """
    Un-transform a user's ratings by adding in the bias model.

    Args:
        user: The user ID.
        ratings(pandas.Series): The user's ratings, indexed by item.
        user_bias(float or None): The user's bias. If `None`, it looks up the user bias learned by `fit`, if available.

    Returns:
        pandas.Series: The user's de-normalized ratings.
    """
```

If the `Bias` is *not* configured to compute user biases, then `transform_user` should return a user bias of 0.

carlos10seg:
Add transform_user and inverse_transform_user methods to bias. Add a new unit test to validate the functionality.

This closes #174.

### DIFF
--- a/lenskit/algorithms/basic.py
+++ b/lenskit/algorithms/basic.py
@@ -153,7 +153,7 @@ class Bias(Predictor):
             rvps['rating'] += rvps['u_off'].fillna(0)
         return rvps.drop(columns=['u_off', 'i_off'])
 
-    def transform_user(self, user, ratings):
+    def transform_user(self, ratings):
         """
         Transform a user's ratings by subtracting the bias model.
 
@@ -165,11 +165,23 @@ class Bias(Predictor):
             pandas.Series:
                 The transformed ratings and the user bias.
         """
-        ratings_df = pd.DataFrame({
-            'rating': ratings.tolist(),
-            'item': ratings.index.tolist(),
-            'user': user})
-        return self.transform(ratings_df)
+        ratings = ratings.subtract(self.mean_)
+
+        # build user_offsets_
+        if self.item_offsets_ is not None:
+            rv = ratings.values - self.item_offsets_[
+                self.item_offsets_.index.isin(ratings.index)].values
+        else:
+            rv = ratings
+        user_offsets_ = self._mean(rv, self.user_damping)
+
+        # subtracts bias
+        if self.item_offsets_ is not None:
+            ratings = ratings.subtract(
+                    self.item_offsets_[self.item_offsets_.index.isin(ratings.index)].values)
+
+        ratings = ratings.subtract(user_offsets_)
+        return ratings, user_offsets_
 
     def inverse_transform_user(self, user, ratings, user_bias=None):
         """
@@ -183,18 +195,17 @@ class Bias(Predictor):
         Returns:
             pandas.Series: The user's de-normalized ratings.
         """
-        rvps = pd.DataFrame({'item': ratings.index.tolist(), 'user': user})
-        rvps['rating'] = ratings.values + self.mean_
-        rvps['rating'].fillna(0, inplace=True)
+        ratings = ratings.add(self.mean_)
         if self.item_offsets_ is not None:
-            rvps = rvps.join(self.item_offsets_, on='item', how='left')
-            rvps['rating'] += rvps['i_off'].fillna(0)
+            ratings = ratings.add(self.item_offsets_[
+                self.item_offsets_.index.isin(ratings.index)].values)
         if user_bias is not None:
-            rvps['rating'] += user_bias
+            ratings = ratings.add(user_bias)
         elif self.user_offsets_ is not None:
-            rvps = rvps.join(self.user_offsets_, on='item', how='left')
-            rvps['rating'] += rvps['u_off'].fillna(0)
-        return rvps.drop(columns=['u_off', 'i_off'])
+            current_u_b = self.user_offsets_[self.user_offsets_.index == user]
+            if len(current_u_b) > 0:
+                ratings = ratings.add(current_u_b.values[0])
+        return ratings
 
     def fit_transform(self, ratings, **kwargs):
         """
--- a/lenskit/algorithms/basic.py
+++ b/lenskit/algorithms/basic.py
@@ -167,21 +167,13 @@ class Bias(Predictor):
         """
         ratings = ratings.subtract(self.mean_)
 
-        # build user_offsets_
-        if self.item_offsets_ is not None:
-            rv = ratings.values - self.item_offsets_[
-                self.item_offsets_.index.isin(ratings.index)].values
-        else:
-            rv = ratings
-        user_offsets_ = self._mean(rv, self.user_damping)
+        ioff = self.item_offsets_.reindex(ratings.index, fill_value=0)
+        ratings = ratings - ioff
 
-        # subtracts bias
-        if self.item_offsets_ is not None:
-            ratings = ratings.subtract(
-                    self.item_offsets_[self.item_offsets_.index.isin(ratings.index)].values)
+        u_offset = self._mean(ratings, self.user_damping)
 
-        ratings = ratings.subtract(user_offsets_)
-        return ratings, user_offsets_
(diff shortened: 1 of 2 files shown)


lenskit/algorithms/basic.py
```diff
@@ -153,6 +153,49 @@ def inverse_transform(self, ratings):
...
+            user: The user ID.
+            ratings(pandas.Series): The user's ratings, indexed by item.
+            user_bias(float or None): If `None`, it looks up the user bias learned by `fit`.
+
+        Returns:
+            pandas.Series: The user's de-normalized ratings.
+        """
+        rvps = pd.DataFrame({'item': ratings.index.tolist(), 'user': user})
+        rvps['rating'] = ratings.values + self.mean_
+        rvps['rating'].fillna(0, inplace=True)
```
mdekstrand, line 188: Can we do all of this with a series, rather than converting to a data frame?

lenskit/algorithms/basic.py
```diff
@@ -153,6 +153,49 @@ def inverse_transform(self, ratings):
...
+
+        Returns:
+            pandas.Series:
+                The transformed ratings and the user bias.
+        """
+        ratings_df = pd.DataFrame({
+            'rating': ratings.tolist(),
+            'item': ratings.index.tolist(),
+            'user': user})
+        return self.transform(ratings_df)
```
mdekstrand, line 172: A few things we need to change here:

- It should return a series, not a data frame
- It should compute the user's offset from *this* set of ratings
- It should return both the ratings and the user bias: `return ratings, u_bias` (this is so we have it available in to pass to `inverse_transform_user`).

tests/test_bias.py
```diff
@@ -280,6 +280,18 @@ def test_bias_separate_damping():
...
+    algo.fit(simple_df)
+
+    new_ratings = pd.Series([4.0, 5.0], index=[1, 2]) # items as index and ratings as values
+    ratings_with_bias = algo.transform_user(13, new_ratings)
+
+    new_ratings_with_bias = ratings_with_bias.drop(columns=['user']).set_index('item')
+    ratings_without_bias = algo.inverse_transform_user(13, new_ratings_with_bias)
+
+    no_bias_new_ratings = ratings_without_bias.drop(columns=['user']).set_index('item').values.flatten()
+    assert no_bias_new_ratings[0] == new_ratings.values[0] and no_bias_new_ratings[1] == new_ratings.values[1]
```
mdekstrand, line 294: We need a test for the inverse_transform taking a bias as well, and then this should be good :).

Can you simplify, though? That will probably happen naturally when you make the bias methods use a series instead of a data frame.

lenskit/algorithms/basic.py
carlos10seg: If local user_offsets is calculated in transform_user and the one created in fit is not used, then, what should I use  in inverse_transform_user method when user_bias=None? The self.user_offsets_ or calculate a new one?

mdekstrand: There is no basis for calculating a new one. Use `self.user_offsets_` if available, otherwise use 0.

lenskit/algorithms/basic.py
```diff
@@ -153,6 +153,60 @@ def inverse_transform(self, ratings):
...
+        """
+        ratings = ratings.subtract(self.mean_)
+
+        # build user_offsets_
+        if self.item_offsets_ is not None:
+            rv = ratings.values - self.item_offsets_[
+                self.item_offsets_.index.isin(ratings.index)].values
+        else:
+            rv = ratings
+        user_offsets_ = self._mean(rv, self.user_damping)
```
mdekstrand, line 176: Since this is only one offset, and it isn't being saved as a field, just call it `u_offset`.

lenskit/algorithms/basic.py
```diff
@@ -153,6 +153,60 @@ def inverse_transform(self, ratings):
...
+        Returns:
+            pandas.Series:
+                The transformed ratings and the user bias.
+        """
+        ratings = ratings.subtract(self.mean_)
+
+        # build user_offsets_
+        if self.item_offsets_ is not None:
+            rv = ratings.values - self.item_offsets_[
+                self.item_offsets_.index.isin(ratings.index)].values
```
mdekstrand, line 173: I don't know that this code is correct - it doesn't do anything to ensure the two arrays are in the same order.

Consider:

```python
ioff = self.item_offsets_.reindex(ratings.index, fill_value=0)
rv = ratings - ioff
```
```

File: lenskit/algorithms/basic.py [hunk 6]
```diff
@@ -153,6 +153,60 @@ def inverse_transform(self, ratings):
...
+            ratings(pandas.Series): The user's ratings, indexed by item.
+            user_bias(float or None): If `None`, it looks up the user bias learned by `fit`.
+
+        Returns:
+            pandas.Series: The user's de-normalized ratings.
+        """
+        ratings = ratings.add(self.mean_)
+        if self.item_offsets_ is not None:
+            ratings = ratings.add(self.item_offsets_[
+                self.item_offsets_.index.isin(ratings.index)].values)
```
mdekstrand, line 201: Same as the `transform_user` comment.

lenskit/algorithms/basic.py
```diff
@@ -153,6 +153,60 @@ def inverse_transform(self, ratings):
...
+        if user_bias is not None:
+            ratings = ratings.add(user_bias)
+        elif self.user_offsets_ is not None:
+            current_u_b = self.user_offsets_[self.user_offsets_.index == user]
```
mdekstrand, line 205: You can do `self.user_offsets_.loc[user]` and it will be faster:

```python
elif self.user_offsets_ is not None and user in self.user_offsets_.index:
    ratings = ratings + self.user_offsets_.loc[user]
```

tests/test_bias.py
```diff
@@ -280,6 +280,35 @@ def test_bias_separate_damping():
...
+    ratings_without_bias = algo.inverse_transform_user(user, new_ratings)
+
+    r_no_b_0 = ratings_without_bias[2]
+    r_u_i_m_0 = new_ratings[2] + algo.user_offsets_.loc[user] + algo.item_offsets_.loc[2] + algo.mean_
+
+    r_no_b_1 = ratings_without_bias[3]
+    r_u_i_m_1 = new_ratings[3] + algo.user_offsets_.loc[user] + algo.item_offsets_.loc[3] + algo.mean_
+
+    assert r_no_b_0 == r_u_i_m_0
+    assert r_no_b_1 == r_u_i_m_1
```
mdekstrand, line 311: let's write these as:

```python
assert ratings_without_bias[2] == new_ratings[2] + algo.user_offsets_.loc[user] + algo.item_offsets_.loc[2] + algo.mean_
```

If it's too long, we can rename `ratings_without_bias` to `result` or `v` or something.

That way, if the assertion fails, PyTest will give us more information about what went wrong.

lenskit/algorithms/basic.py
```diff
@@ -153,6 +153,51 @@ def inverse_transform(self, ratings):
...
+            ratings(pandas.Series): The user's ratings, indexed by item.
+                Must have at least `item` as index and `rating` column.
+
+        Returns:
+            pandas.Series:
+                The transformed ratings and the user bias.
+        """
+        ratings = ratings.subtract(self.mean_)
+
+        ioff = self.item_offsets_.reindex(ratings.index, fill_value=0)
```
mdekstrand, line 170: I think we have to deal with the (unusual) case when `item_offsets` is None. I think we can protect these lines with

    if self.item_offsets_ is not None:

and do the same in `inverse_transform_user`.

Issue #175, opened 2020-06-19
Move Bias class


mdekstrand:
The Bias class is getting pretty big, and `basic` is a grab-bag. We should:

- [ ] Move `Bias` into a new module `lenskit.algorithms.bias`
- [ ] Make `lenskit.algorithms.basic.Bias` an alias for compatibility
- [ ] Update documentation links

I would like to do this with #174.

PR #183, opened 2020-08-12
Move Bias class into `bias` package (#175)

carlos10seg:
Changes for #175


doc/impl-tips.rst
```diff
@@ -49,7 +49,7 @@ the derived structures so they can also be shared.
 
 **If your algorithm uses subsidiary models as a part of the training process**, but does not need them
 for prediction or recommendation (for example, :py:class:`lenskit.algorithms.als.BiasMF`'s use of
-:py:class:`lenskit.algorithms.basic.Bias` in ``fit``, during which it copies the bias model's
+:py:class:`lenskit.algorithms.bias.Bias` in ``fit``, during which it copies the bias model's
```
mdekstrand, line 52: Ooh, we need to update these docs - this isn't how it works any more, since #173.

doc/basic.rst
```diff
@@ -1,14 +1,12 @@
 Basic and Utility Algorithms
 ============================
 
-.. module:: lenskit.algorithms.basic
-
-The :py:mod:`lenskit.algorithms.basic` module contains baseline and utility algorithms
+The :py:mod:`lenskit.algorithms.basic` and :py:mod:`lenskit.algorithms.bias` modules contain baseline and utility algorithms
```
mdekstrand, line 4: I think we should move the `Bias` docs to their own page, `bias.rst`, that should probably go after the `basic.rst` page in the algorithm documentation list.

lenskit/algorithms/basic.py
```diff
@@ -12,255 +12,9 @@
...
-                The transformed ratings and the user bias.
-        """
-        ratings = ratings.subtract(self.mean_)
-
-        if self.item_offsets_ is not None:
-            ioff = self.item_offsets_.reindex(ratings.index, fill_value=0)
-            ratings = ratings - ioff
-
-        u_offset = self._mean(ratings, self.user_damping)
+from .bias import Bias
```
mdekstrand, line 15: good! Exactly what we need to do.

lenskit/algorithms/bias.py
```diff
@@ -0,0 +1,254 @@
+import logging
+
+import pandas as pd
+
+from . import Predictor
+
+_logger = logging.getLogger(__name__)
+
+class Bias(Predictor):
```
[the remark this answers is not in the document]
mdekstrand: this looks like it might be fixed? not sure why it's still flagged here.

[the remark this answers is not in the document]
carlos10seg: I think, it is already fixed.
image

doc/basic.rst
carlos10seg: Is this the desired text and format for bias.rst?
image

mdekstrand: Yep!

mdekstrand: We want the lead paragraph to talk about it just providing the bias model, rather than general baseline and utility algorithms, but yes, it's correct.

We can also delete the Personalized Mean... header.

Issue #182, opened 2020-08-11
Create a known-rating predictor


mdekstrand:
We need a known-rating predictor that memorizes the ratings provided to `fit`, and in `predict` just returns those ratings.

This will be very similar to `Memorized`. We may want to just adjust `Memorized` to learn in `fit` if no ratings were provided to the constructor.

mdekstrand:
Let's **not** modify Memorized. Make a new one.

PR #184, opened 2020-08-17
Add known-rating predictor (#182)

carlos10seg:
PR for #182


mdekstrand:
Looks like I'll have to review this after we merge #183 - right now the changes view shows too many changes.

lenskit/algorithms/basic.py
```diff
@@ -538,3 +292,18 @@ def recommend(self, user, n=None, candidates=None, ratings=None):
...
         return 'Random'
+
+
+class KnownRating(Predictor):
+    """
+    The known rating algorithm memorizes ratings provided in the fit method.
+    """
+
+    def fit(self, ratings, **kwargs):
+        self.ratings = ratings
```
mdekstrand, line 303: We should consider setting up indexes so this in a consistent state, and then using them in `predict_for_user`.

lenskit/algorithms/basic.py
```diff
@@ -538,3 +292,18 @@ def recommend(self, user, n=None, candidates=None, ratings=None):
...
+        return self
+
+    def predict_for_user(self, user, items, ratings=None):
+        uscores = self.ratings[self.ratings.user == user]
```
mdekstrand, line 307: If we set up `ratings` to have an index in `fit`, then this can use an index for better performance on large data sets.

lenskit/algorithms/basic.py
mdekstrand: A `MultiIndex` on `[user, item]` will do the trick, then sort by index (`sort_index`). This class is simple enough we won't have the problems mutli-indexes usually have.

lenskit/algorithms/basic.py
```diff
@@ -538,3 +292,18 @@ def recommend(self, user, n=None, candidates=None, ratings=None):
...
+        urates = uscores.set_index('item').rating
+        return urates.reindex(items)
```
mdekstrand, line 309: Exactly the right way to do this.

PR #196, opened 2020-10-09
[DOC/FIX] Correction in ImplicitMF docstring

ShwetanshuSingh:
Update default input value for `method` parameter in the docstring of ImplicitMF


mdekstrand:
Thank you for this!

PR #214, opened 2020-12-27
Allow scipy='coo' in sparse_ratings

### LINKED ISSUE #213: Support scipy.sparse.coo_matrix in sparse_ratings
It would be good for `sparse_ratings` to support creating a `scipy.sparse.coo_matrix` directly, so integrations can use it. As it is, it returns `scipy.sparse.csr_matrix` with `scipy=True`. This can be enabled with an additional keyword argument without breaking existing usage of the API.

(This is needed for the lightfm integration in issue #188; as it is, we either have to perform an additional conversion after calling `sparse_ratings` or substantially reproduce its functionality).

reppertj:
Return a `scipy.sparse.coo_matrix` from `sparse_ratings` via the argument `scipy='coo'` (Fixes #213).

### DIFF
--- a/lenskit/matrix/__init__.py
+++ b/lenskit/matrix/__init__.py
@@ -11,7 +11,7 @@ import scipy.sparse as sps
 
 _logger = logging.getLogger(__name__)
 
-RatingMatrix = namedtuple('RatingMatrix', ['matrix', 'users', 'items'])
+RatingMatrix = namedtuple("RatingMatrix", ["matrix", "users", "items"])
 RatingMatrix.__doc__ = """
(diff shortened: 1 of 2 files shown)


mdekstrand:
Code looks good to me. I'd like to understand why tests failed on Mac with OpenBLAS - triggering a re-build to see if it's ephemeral.

mdekstrand:
Upon inspection of the test suite failure, I believe it is unrelated - something perhaps changed in an OpenBLAS update to cause a subtle failure on Mac. I am re-running the `master` tests to confirm.

If that is the case, we'll be good to merge this change even with those tests failing.

PR #369, opened 2024-03-11
Documentation fix for Recall and Hit

lukas-wegmeth:
The documentation contains max rather than min functions. Min functions are used in the implementation.

### DIFF
--- a/lenskit/metrics/topn.py
+++ b/lenskit/metrics/topn.py
@@ -74,7 +74,7 @@ def recall(recs, truth, k=None):
     Compute recommendation recall.  This is computed as:
 
     .. math::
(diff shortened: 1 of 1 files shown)


lukas-wegmeth:
Could also remove nrel variable from the hit function and remove the formula from the documentation as the metric is more concisely explained in its description.

mdekstrand:
Good catch — thank you!

mdekstrand:
The code change is causing test failures — need to page in my memory of what's happening there to figure out why.

lukas-wegmeth:
My bad! Ran the test suite on the wrong branch. Fix is coming as soon as tests pass.

PR #408, opened 2024-05-29
Add Torch matrix test + improve sparse matrix hypothesis infra

mdekstrand:
This improves the test utilities for generating sparse matrices for testing, and adds tests for PyTorch sparse matrix logic to ensure that our underlying compute substrate is well-behaved for algorithm implementation.


mdekstrand:
This is testing PyTorch bug pytorch/pytorch#127491, which is exercised in isolation in https://github.com/mdekstrand/torch-spmv-bug-repro.

Issue #385, opened 2024-05-06
Add ability to load algorithms from TOML files


mdekstrand:
It would be useful to be able to load and configure algorithms from serialized files (e.g. TOML).

This will enable general command-line tools.

mdekstrand:
This is supported by the pipeline configuration capabilities. The configuration files aren't very ergonomic, but it works.

PR #469, opened 2024-08-07
Serialize full pipeline configurations

mdekstrand:
This extends the configuration support to enable entire pipeline configurations to be serialized and deserialized. It is also a prerequisite for implementing #385. This version of the changes only generates the serialization — it does not save or load them to files, that will be a separate change.

It also renames and refactors some pipeline test files for clarity and improves error handling.

Working:

- [x] Input nodes (and their types)
- [x] Component nodes
- [x] Component wirings
- [x] Fallback nodes
- [x] Literal nodes
- [x] Pipeline configuration hashing
- [x] Pipeline metadata

### DIFF
--- a/lenskit/lenskit/pipeline/__init__.py
+++ b/lenskit/lenskit/pipeline/__init__.py
@@ -493,7 +493,7 @@ class Pipeline:
         dependencies will produce the same hash.  In LensKit 2024.1, the
         configuration hash is computed by computing the JSON serialization of
         the pipeline configuration *without* a hash returning the hex-encoded
-        SHA1 hash of that configuration.
+        SHA256 hash of that configuration.
         """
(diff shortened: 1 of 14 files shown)


lenskit/lenskit/pipeline/__init__.py
```diff
@@ -139,7 +143,7 @@ def create_input(self, name: str, *types: type[T] | None) -> Node[T]:
         """
         self._check_available_name(name)
 
-        node = InputNode[Any](name, types=set((t if t is not None else type[None]) for t in types))
+        node = InputNode[Any](name, types=set((t if t is not None else type(None)) for t in types))
```
mdekstrand, line 146: Bugfix for existing code around `None`.

PR #481, opened 2024-09-11
fix: invalid file references to requirements files

ZiyaoWei:
Fixed a bug where `requirements-full-dev.txt` seems to be using an outdated reference to dev-requirements.txt by changing it to requirements-dev.txt instead. Also fixed a similar bug in README.md for uv builds.


mdekstrand:
Thanks!

Issue #359, opened 2024-02-21
Support time-bounded popularity


mdekstrand:
Right now, `Popular` is based on overall popularity.

We should add support for *time-bounded* popularity: most popular items in the last N days.

Time-weighted popularity could also be interesting.

mdekstrand:
Hi @ZiyaoWei — very glad you're interested in helping! You're right there's a lot of activity, and it makes things a bit of a moving target (and having built the pipeline and dataset abstractions, the next big lift there is to overhaul all of the model implementations to use them — so `PopScore` and friends are changing). I'm also on vacation this week, and will be returning to development on Aug. 26. I think this is a good first issue, though, and would love to provide what help you would need to contribute.

I think having it be a separate class from `PopScore` might make sense, just to keep `PopScore` simple.

If you wanted to work on a PR for the current APIs, you'd be welcome to; if you want to wait a couple weeks or so for the model API overhaul to land and do itagains the new code, that would remove the need to learn how to code for the current API.

mdekstrand:
@ZiyaoWei The most important pieces for this bit are actually there now — it isn't hooked into the evaluator yet, but the new pipeline component `PopScore` is in `lenskit.basic`, so you can get started on this if you want!

mdekstrand:
Yeah, let's make that a separate change (and probably a separate component). There are a few models in the literature but nothing standard.

mdekstrand:
Yes, that sounds like a good idea!

PR #493, opened 2024-10-18
feat: Add TimeBoundedPopScore for time-bounded popularity

ZiyaoWei:
Implements #359.

### DIFF
--- a/lenskit/lenskit/algorithms/basic.py
+++ b/lenskit/lenskit/algorithms/basic.py
@@ -10,7 +10,6 @@ Basic utility algorithms and combiners.
 
 import logging
 from collections.abc import Iterable, Sequence
-from datetime import datetime
 from typing import overload
 
 import numpy as np
@@ -42,7 +41,7 @@ class PopScore(Predictor):
             - ``'count'``
 
     Attributes:
-        item_scores_(pandas.Series):
+        item_pop_(pandas.Series):
             Item popularity scores.
     """
 
@@ -52,13 +51,8 @@ class PopScore(Predictor):
     @override
     def fit(self, data: Dataset, **kwargs):
         _logger.info("counting item popularity")
-
-        counts = data.item_stats()["count"]
-        self.item_scores_ = self._fit_internal(counts, **kwargs)
-
-        return self
-
-    def _fit_internal(self, scores: pd.Series, **kwargs):
+        stats = data.item_stats()
+        scores = stats["count"]
         if self.score_method == "rank":
             _logger.info("ranking %d items", len(scores))
             scores = scores.rank().sort_index()
@@ -74,7 +68,9 @@ class PopScore(Predictor):
         else:
             raise ValueError("invalid scoring method " + repr(self.score_method))
 
-        return scores
+        self.item_scores_ = scores
+
+        return self
 
     @override
     def predict_for_user(self, user, items, ratings=None):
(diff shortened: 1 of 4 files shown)


lenskit/lenskit/algorithms/basic.py
```diff
@@ -80,6 +84,55 @@ def __str__(self):
...
+            - ``'quantile'`` (the default)
+            - ``'rank'``
+            - ``'count'``
+
+    Attributes:
+        item_scores_(pandas.Series):
+            Time-bounded item popularity scores.
+    """
+
+    def __init__(self, time_window: timedelta, score_method="quantile"):
```
mdekstrand, line 108: Specifying this as a time delta introduces a behavioral dependency on the current time when the data is fit, which seems problematic for consistency and reproducibility. The same call with the same inputs will select different data based on when it is run.

I think it would probably be better to specify this as a cutoff time (which can then be set to predictably simulate a variety of times in an experiment).

(An alternative would be to specify it as a time delta that is computed from when the query is scored — queries will eventually get support for "query time" for offline experiments — instead from when the model is trained, but that would involve re-computing scores at query time which would be expensive).

lenskit/lenskit/algorithms/basic.py
```diff
@@ -80,6 +84,55 @@ def __str__(self):
...
+
+    @override
+    def fit(self, data: Dataset, **kwargs):
+        _logger.info("counting time-bounded item popularity")
+
+        log = data.interaction_log("numpy")
+
+        counts = np.zeros(data.item_count, dtype=np.int32)
+        start_timestamp = (datetime.now() - self.time_window).timestamp()
+        item_nums = log.item_nums[log.timestamps is not None and log.timestamps > start_timestamp]
```
mdekstrand, line 122: I'm not sure what `log.timestamps is not None` will do here when it's actually None (we'll have a single value of `False` as the index). Better to put the `none` check outside of the indexing — if timestamps is `None`, then issue a warning (because time-bounded popularity without timestamps is meaningless) and fall back to the unbounded popularity.

lenskit/tests/algorithms/test_time_bounded_popular.py
```diff
@@ -0,0 +1,61 @@
...
+def test_time_bounded_pop_score_quantile_two_day_window():
+    algo = basic.TimeBoundedPopScore(2 * day)
+    algo.fit(simple_ds)
+    assert algo.item_scores_.equals(pd.Series([0.25, 1.0, 0.5], index=[1, 2, 3]))
+
+
+def test_time_bounded_pop_score_rank():
+    algo = basic.TimeBoundedPopScore(2 * day, "rank")
+    algo.fit(simple_ds)
+    assert algo.item_scores_.equals(pd.Series([1.5, 3.0, 1.5], index=[1, 2, 3]))
```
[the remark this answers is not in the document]
mdekstrand: Yes, low-level for these component-level unit tests are fine, although we should have some tests that exercise the `__call__` interface.

mdekstrand (changes requested):
Thank you very much for submitting this! I've left some comments; in particular, I think we should rethink the configuration interface a bit (specify a cutoff timestamp instead of a time delta).

mdekstrand:
Sorry for the delay in second review, been busy over here. Will look this week.

mdekstrand:
Merged, thank you very much @ZiyaoWei!

PR #621, opened 2025-02-04
[MINOR:fix link] Update README.md

cakiki:
Getting started link doesn't resolve on /latest.

There also seems to be a general issue with the /latest branch since it shows a completely different documentation page than the one on readthedocs. (see: https://lkpy.lenskit.org/en/latest vs https://lkpy.lenskit.org/en/stable/index.html)


mdekstrand:
Thanks! Yes, the new branch is has rearranged documents, to hopefully have a better organization.

PR #671, opened 2025-03-21
adding again lightfm

danhdanhtuan0308:


tests/lightfm/test_lenskit_lightfm.py
```diff
@@ -0,0 +1,99 @@
+import logging
+import pickle
+
+import numpy as np
+import pandas as pd
+from pytest import approx, mark
+import sys 
+import os 
+sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
```
mdekstrand, line 9: Please remove — we do not need to do path manipulations for imports. If there is an import problem, we should fix it at the source or in how we set up and run things.

tests/lightfm/test_lenskit_lightfm.py
```diff
@@ -0,0 +1,99 @@
...
+
+from lenskit.data import Dataset, ItemList, from_interactions_df
+from lenskit.metrics import quick_measure_model
+from lenskit.testing import BasicComponentTests, ScorerTests
+from lenskit_lightfm import LightFMScorer
```
mdekstrand, line 14: changing this to `from lenskit.lightfm` should fix the immediate test failure problem.

src/lenskit/lightfm.py
```diff
@@ -0,0 +1,119 @@
...
+            Additional arguments to pass to the LightFM constructor.
+    """
+
+    config: LightFMConfig
+
+    users_: Vocabulary
+    items_: Vocabulary
+    user_features_: np.ndarray[tuple[int, int], np.dtype[np.float64]]
+    item_features_: np.ndarray[tuple[int, int], np.dtype[np.float64]]
+    model_: LightFM
```
mdekstrand, line 67: Let's remove the `_` suffixes on these — with the new design, I have not been using that suffix in new components.

src/lenskit/lightfm.py
```diff
@@ -0,0 +1,119 @@
...
+
+    @override
+    def train(self, data: Dataset, options: TrainingOptions = TrainingOptions()):
+        if hasattr(self, "model_") and not options.retrain:
+            return
+
+        interaction_matrix = data.interaction_matrix(format="csr", field="rating")
```
mdekstrand, line 74: This will only work for explicit-feedback data, but the defaults are set up for WARP loss.

If LightFM supports an explicit-feedback mode, we should only use the `rating` column if we are using that mode. It would be better to supply no field, so we get a binary indicator matrix, for the common implicit feedback case.

src/lenskit/lightfm.py
```diff
@@ -0,0 +1,119 @@
...
+        kwargs:
+            Additional arguments to pass to the LightFM constructor.
+    """
+
+    config: LightFMConfig
+
+    users_: Vocabulary
+    items_: Vocabulary
+    user_features_: np.ndarray[tuple[int, int], np.dtype[np.float64]]
+    item_features_: np.ndarray[tuple[int, int], np.dtype[np.float64]]
```
mdekstrand, line 66: Since we are using the model to do predictions, we don't need to save these arrays separately, so let's remove them.

src/lenskit/lightfm.py
```diff
@@ -0,0 +1,119 @@
...
+    def __call__(self, query: QueryInput, items: ItemList) -> ItemList:
+        query = RecQuery.create(query)
+
+        user_id = query.user_id
+        user_num = None
+        if user_id is not None:
+            user_num = self.users_.number(user_id, missing=None)
+        if user_num is None:
+            _logger.debug("Unknown user %s", query.user_id)
+            return ItemList(items, scores=np.nan)
```
mdekstrand, line 103: Does LightFM have the ability to predict for a user based on their history instead of their identifier? If so, we should use that functionality.

tests/lightfm/test_lenskit_lightfm.py
```diff
@@ -0,0 +1,99 @@
...
+    preds2 = new_algo(10, ItemList([3])).scores("pandas", index="ids")
+    assert np.allclose(preds1, preds2, equal_nan=True)
+
+
+@mark.slow
+@mark.eval
+def test_lightfm_batch_accuracy(ml_100k: pd.DataFrame):
+    data = from_interactions_df(ml_100k)
+    lightfm_algo = LightFMScorer()
+    results = quick_measure_model(lightfm_algo, data, predicts_ratings=True)
```
mdekstrand, line 96: The default setup is WARP loss, which is for implicit feedback. So we should remove `predicts_ratings`, and change to top-N metrics (RBP or NDCG).

mdekstrand (changes requested):
Great start! Need to get the tests working, and I left some other comments for removing unnecessary bits and improving the logic.

PR #676, opened 2025-03-24
Limit test users to only have users who had item in the train data.

sushobhan2024:


src/lenskit/splitting/temporal.py
```diff
@@ -94,6 +99,10 @@ def split_global_time(
             tlog.debug("filtering test data for upper bound")
             test = matrix[mask & (ts_col < t2)]
 
+        if filter_test_users:
+            train_users = set(train_build.build().interactions().pandas(ids=True)["user_id"])
+            test = test[test["user_id"].isin(train_users)]
+
         tlog.debug("building training data set")
         train_ds = train_build.build()
```
mdekstrand, line 107: Should move the `train_ds = ` line before the `if` and use that variable in the filtering, so we only build the data set once.

src/lenskit/splitting/temporal.py
```diff
@@ -94,6 +99,10 @@ def split_global_time(
             tlog.debug("filtering test data for upper bound")
             test = matrix[mask & (ts_col < t2)]
 
+        if filter_test_users:
+            train_users = set(train_build.build().interactions().pandas(ids=True)["user_id"])
```
mdekstrand, line 103: This will be more efficient if we keep things in Pandas / NumPy data structures, and don't need to de-duplicate.

What we can do is get the users from the user statistics — the `index` of the user statistics data frame has the user IDs:

```python
user_data = train_ds.user_stats()
train_users = user_data.index[user_data['count'] > 0]
```

tests/splitting/test_split_temporal_filter.py
```diff
@@ -0,0 +1,34 @@
...
+    point = dt.datetime.fromisoformat("2015-01-01")
+
+    split = split_global_time(ml_ds, "2015-01-01", filter_test_users=True)
+    assert np.all(split.train.interaction_table(format="pandas")["timestamp"] < point)
+
+    for u, il in split.test:
+        ts = il.field("timestamp", format="pandas")
+
+        assert ts is not None
+        assert np.all(ts >= point)
```
mdekstrand, line 27: we also need to verify that all of the test users have at least one training rating.

mdekstrand (changes requested):
Looks pretty good! A couple of efficiency comments, and one test correctness change we need to make.

src/lenskit/splitting/temporal.py
sushobhan2024: If we filter this way, we don't need to use the variable `train_ds` in filtering, right?

sushobhan2024: Correct!

mdekstrand (approved):

PR #683, opened 2025-03-27
run Doctor under coverage

mdekstrand:


mdekstrand:
Well that didn't seem to do anything.

PR #739, opened 2025-04-29
Adding Mean Average Precision

albus-droid:
1. I created a module named map inside src/lenskit/metrics/ranking
2. It returns the average precision of a single user
3. I also added some simple test cases


albus-droid:
I made the requested changes. Let me know if there is anything more.

and also Thank you for letting me work on this.

mdekstrand (approved):

mdekstrand:
@albus-droid thank you for doing it!

PR #741, opened 2025-04-29
Add sklearn NMF as a scorer

### LINKED ISSUE #686: Add non-negative matrix factorization (NMF)
Since scikit-learn provides `NMF` (non-negative matrix factorization), it would be easy to add this as an implicit-feedback recommendation model.

FroggoLight:
Added non_negative_factorization from sklearn

Closes #686.

### DIFF
--- a/src/lenskit/sklearn/nmf.py
+++ b/src/lenskit/sklearn/nmf.py
@@ -5,7 +5,7 @@
 # SPDX-License-Identifier: MIT
 
 """
-This is for implicit 
+This is for implicit
 
 This module contains a truncated SVD explicit-feedback scorer built on
 :class:`sklearn.decomposition.TruncatedSVD`.
@@ -13,12 +13,11 @@ This module contains a truncated SVD explicit-feedback scorer built on
 
 from __future__ import annotations
 
-from dataclasses import dataclass
-
 import numpy as np
 from numpy.typing import NDArray
+from pydantic import BaseModel
 from sklearn.decomposition import non_negative_factorization
-from typing_extensions import Literal, override, Union
+from typing_extensions import Literal, override
 
 from lenskit.data import Dataset, ItemList, QueryInput, RecQuery
 from lenskit.data.vocab import Vocabulary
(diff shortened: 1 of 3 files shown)


src/lenskit/sklearn/nmf.py
```diff
@@ -0,0 +1,99 @@
...
+    l1_ratio: float = 0.0
+
+
+class NMFScorer(Component[ItemList], Trainable):
+    config: NMFConfig
+
+    users_: Vocabulary
+    items_: Vocabulary
+    user_components_: NDArray[np.float64]
+    item_components_: NDArray[np.float64]
```
mdekstrand, line 46: 1. In newer code, I have been omitting the trailing `_`, and naming the component / embedding matrices `user_embeddings` and `item_embeddings`. So let's changes these (both the Vocabulary and NDArray attribute names) for consistency.
2. We have an `NPMatrix` type that is more precise than `NDArray`, let's use that instead as the type of the component declarations.
3. Let's use `np.float32`; you will need to use `np.require` in `train` to convert the component matrices to float32. This will save space in the final models (and 64-bit precision usually isn't needed anyway).

tests/sklearn/test_nmf.py
```diff
@@ -0,0 +1,103 @@
...
+def test_nmf_save_load(ml_ds: Dataset):
+    original = nmf.NMFScorer(n_components=20)
+    original.train(ml_ds)
+
+    mod = pickle.dumps(original)
+    _log.info("serialized to %d bytes", len(mod))
+    algo = pickle.loads(mod)
+
+    assert np.all(algo.user_components_ == original.user_components_)
+    assert np.all(algo.item_components_ == original.item_components_)
```
mdekstrand, line 91: Let's drop this test. Pickling is already tested by the common component tests, and this model trains slowly.

tests/sklearn/test_nmf.py
```diff
@@ -0,0 +1,103 @@
...
+@mark.slow
+@mark.eval
+def test_nmf_batch_accuracy(rng, ml_100k: pd.DataFrame):
+    data = from_interactions_df(ml_100k)
+    svd_algo = nmf.NMFScorer(n_components=25)
+    results = quick_measure_model(svd_algo, data, predicts_ratings=True, rng=rng)
+
+    ndcg = results.list_summary().loc["NDCG", "mean"]
+    _log.info("nDCG for users is %.4f", ndcg)
+    assert ndcg > 0.22  # type: ignore
```
mdekstrand, line 103: possible TODO, but not for this ticket: refactoring so this is part of the standard suite instead of rewriting basically the same test for every different component.

maybe that's your next ticket?

mdekstrand (changes requested):
Overall code looks good. Some small changes, and then the big one: we need documentation. We need to write docstrings for both the scorer and config classes, describing what the model does and providing links to the documentation for more. You can see SVD for an example of this.

mdekstrand (approved):

PR #787, opened 2025-06-03
Add Bag of Little Bootstraps

### LINKED ISSUE #785: Implement Bag of Little Bootstraps
To support statistical analysis (#784) and distributional analysis (#414), it would be useful to support the [Bag of Little Bootstraps](http://arxiv.org/abs/1112.5016). Since there does not seem to be an implementation in any of the standard Python statistics packages, we should implement one under `lenskit.stats`.

The first implementation should be sequential. A parallel implementation with Ray (or with free-threaded Python) would be a good idea, but we can do sequential first.

mdekstrand:
Sequential Bag of Little Bootstraps that seems to work, with some analysis of its convergence.

Closes #785.


src/lenskit/stats/_blb.py
```diff
@@ -0,0 +1,491 @@
...
+    """
+    Callable interface for weighted statistics, required by the Bag of Little Bootstraps.
+    """
+
+    def __call__(
+        self,
+        a: NDArray[np.floating[Any]],
+        /,
+        *,
+        weights: NDArray[np.floating[Any] | np.integer[Any]] | None = None,
```
kluver, line 46: I think BLB weights are strictly integer? Not that it matters much.

src/lenskit/stats/_blb.py
```diff
@@ -0,0 +1,491 @@
...
+
+    Args:
+        xs:
+            The array of values to summarize.
+        stat:
+            The statistic to compute.  The Bag of Little Bootstraps requires
+            statistics to support weighted computation (this is what allows it
+            to speed up the bootstrap procedure).
+        ci_width:
+            The width of the confidence interval to estimat.e
```
kluver: typo

kluver:
```suggestion
            The width of the confidence interval to estimate.
```

src/lenskit/stats/_blb.py
```diff
@@ -0,0 +1,491 @@
...
+            The RNG or seed for randomization.
+
+    Returns:
+        A dictionary of statistical results of the statistic.
+    """
+    if stat != "mean":
+        raise ValueError(f"unsupported statistic {stat}")
+
+    mask = np.isfinite(xs)
+    if ninf := int(np.sum(~mask)):
```
kluver: I'm skeptical that `:=` is making this code more readable. But I'm also happy to have learned a new bit of syntax.

src/lenskit/stats/_blb.py
```diff
@@ -0,0 +1,491 @@
...
+            means.record(stat)
+            vars.record(stat)
+
+            stats = means.values
+            # ql, qh = _bca_range(estimate, stats, self.config.ci_margin, accel)
+            # self._tracer.trace("bias-corrected quantiles: [%.4f, %.4f]", ql, qh, accel=accel)
+            ql, qh = ci_quantiles(self.config.ci_width)
+            lb, ub = np.quantile(stats, [ql, qh])
+            self._tracer.trace("initial bounds: %f < s < %f", lb, ub)
+            # recenter bounds around estimate
```
kluver: This _feels_ wrong to me.

mdekstrand: yep we found a bug here. (I'm still trying to find where I found that recenter bounds trick.)

src/lenskit/stats/_blb.py
mdekstrand: yeah idk that it's really helping here.

src/lenskit/stats/_blb.py
mdekstrand: Yeah, doesn't matter a lot here.

PR #792, opened 2025-06-12
Unify list and decomposed metric interfaces

sushobhan2024:
- [x] Moved decomposed metric methods into the main `Metric` class, renamed for clarity
- [x] Adjusted `summarize` method interface for broader input support
- [x] Added default `summarize` and `extract_list_metrics` implementations in `ListMetric`
- [x] Kept `DecomposedMetric` as deprecated wrapper for compatibility
- [x] Added a class `MetricAccumulator` that includes `add_metric`, `measure_list`, `list_metrics`, and `summary_metrics`.
- [x] Modify `RunAnalysis` to use `MetricAccumulator`.
- [x] Write tests.

### DIFF
--- a/src/lenskit/metrics/_accum.py
+++ b/src/lenskit/metrics/_accum.py
@@ -196,14 +196,14 @@ class MetricAccumulator:
 
         return df
 
-    def summary_metrics(self) -> pd.DataFrame:
+    def summary_metrics(self) -> dict[str, float]:
         """
         Compute summary statistics by calling each metric's summarize() method.
 
         Returns:
-            A DataFrame indexed by metric label.
+            A dictionary with flattened metric results.
         """
-        summaries = {}
+        results = {}
 
         for wrapper in self.metrics:
             if wrapper.is_global:
@@ -221,16 +221,12 @@ class MetricAccumulator:
                 summary = wrapper.summarize(intermediate_values)
                 if summary is not None:
                     if isinstance(summary, dict):
-                        summaries[label] = summary
+                        for key, value in summary.items():
+                            results[f"{label}.{key}"] = value
                     else:
-                        summaries[label] = {label: summary}
+                        results[label] = summary
 
-        if not summaries:
-            return pd.DataFrame()
-
-        df = pd.DataFrame.from_dict(summaries, orient="index")
-        df.index.name = "metric"
-        return df
+        return results
 
     def _wrap_metric(
         self,
--- a/src/lenskit/metrics/bulk.py
+++ b/src/lenskit/metrics/bulk.py
@@ -226,15 +226,15 @@ class RunAnalysis:
             if result is not None:
                 global_results[metric_wrapper.label] = result
 
-        summary_df = self._accumulator.summary_metrics()
-        if not summary_df.empty:
-            for metric_name in summary_df.index:
-                if "." not in metric_name and metric_name not in global_results:
-                    for col in summary_df.columns:
-                        value = summary_df.loc[metric_name, col]
-                        if not pd.isna(value):
-                            global_results[metric_name] = value
-                            break
+        summary_results = self._accumulator.summary_metrics()
+        for key, value in summary_results.items():
+            if "." in key:
+                base_metric = key.split(".", 1)[0]
+                if base_metric not in global_results:
+                    global_results[base_metric] = value
+            else:
+                if key not in global_results:
+                    global_results[key] = value
 
         global_results = pd.Series(global_results, dtype=np.float64)
 
--- a/tests/eval/test_metric_accumulator.py
+++ b/tests/eval/test_metric_accumulator.py
@@ -46,17 +46,17 @@ def test_accumulator_empty_and_unmeasured_defaults():
     # initial state
     assert acc.metrics == []
     assert acc.list_metrics().empty
-    assert acc.summary_metrics().empty
+    assert not acc.summary_metrics()
 
     # measuring with no metrics
     acc.measure_list(ItemList([1]), ItemList([1]), user="u1")
     assert acc.list_metrics().empty
-    assert acc.summary_metrics().empty
+    assert not acc.summary_metrics()
 
     # add metrics but do not measure
     acc.add_metric(Recall(5))
     acc.add_metric(NDCG(5))
-    assert acc.summary_metrics().empty
+    assert not acc.summary_metrics()
 
 
 # measuring lists
@@ -74,7 +74,7 @@ def test_accumulator_measures_list_and_summary(sample_lists):
 
     assert len(list_metrics) == 2
     assert set(list_metrics["N"]) == {2.0, 3.0}
-    assert approx(summary.loc["N", "N"]) == 2.5
+    assert approx(summary["N"]) == 2.5
 
 
 def test_accumulator_empty_itemlists():
@@ -210,7 +210,7 @@ def test_scalar_metric():
     acc.add_metric(ScalarMetric())
     acc.measure_list(ItemList([1]), ItemList([1]), user="u1")
     summary = acc.summary_metrics()
-    assert summary.loc["scalar", "scalar"] == 99.0
+    assert summary["scalar"] == 99.0
 
 
 # metrics returning mixed
@@ -242,8 +242,8 @@ def test_mixed_metric_behavior():
     df = acc.list_metrics()
     summary = acc.summary_metrics()
 
-    assert "Mixed" in summary.index
-    assert summary.loc["Mixed", "mean"] == 1.5
+    assert "Mixed.mean" in summary
+    assert summary["Mixed.mean"] == 1.5
     assert len(df) == 2
 
 
@@ -309,7 +309,7 @@ def test_measure_metric_with_none_summarize():
     acc.measure_list(ItemList([1]), ItemList([1]), user="u1")
 
     summary = acc.summary_metrics()
-    assert "no_summarize" not in summary.index
+    assert "no_summarize" not in summary
 
 
 # test with movielens data
@@ -337,12 +337,14 @@ def test_full_workflow_integration_improved(ml_ds):
     assert len(list_metrics) == len(test_users)
     assert set(list_metrics.columns) == {"Recall@10", "NDCG@10"}
     assert all(0 <= list_metrics[col].max() <= 1 for col in list_metrics.columns)
-    assert set(summary.index) == {"Recall@10", "NDCG@10"}
-    assert all(0 <= summary.loc[metric, "mean"] <= 1 for metric in summary.index)
-    for metric in summary.index:
-        assert summary.loc[metric, "mean"] is not None
-        std_val = summary.loc[metric, "std"]
-        assert std_val is None or std_val >= 0
+
+    assert len(summary) > 0
+    for key, value in summary.items():
+        if value is not None:
+            if "std" in key.lower():
+                assert value >= 0
+            else:
+                assert 0 <= value <= 1
 
 
 # test that global metric raises errors for unsupported operations
@@ -415,7 +417,7 @@ def test_empty_intermediate_values():
     acc.add_metric(TestMetric())
 
     summary = acc.summary_metrics()
-    assert summary.empty or "test" not in summary.index
+    assert not summary or "test" not in summary
 
 
 def test_accumulator_duplicate_labels():


src/lenskit/metrics/_base.py
```diff
@@ -72,6 +103,25 @@ def measure_list(self, output: ItemList, test: ItemList, /) -> float:
...
 
+    def extract_list_metrics(self, metric: object, /) -> float | dict[str, float] | None:
+        """
+        Return the given per-list metric result.
+        """
+        return metric
+
+    def summarize(self, values: list[object] | np.ndarray | pa.Array | pa.ChunkedArray, /) -> float:
+        """
+        Aggregarte per-list metric values by computing their mean, ignoring NaNs.
```
mdekstrand: We should define methods primarily in terms of their interface. The _implementation_ computes mean, but that is not a guarantee of the overall contract.

I think perhaps the clearest way to handle this is to remove the docstring from the method implementation, and note in the class's docstring that it provides a default implementation of the `summarize` method using mean.

src/lenskit/metrics/_base.py
```diff
@@ -43,6 +46,34 @@ def label(self) -> str:
     def __str__(self):
         return f"Metric {self.label}"
 
+    @abstractmethod
+    def measure_list(self, output: ItemList, test: ItemList, /) -> object:
+        """
+        Compute measurements for a single list.
```
mdekstrand, line 65: We should define what the return data is and how it is intended to be used, and why it might be something other than a `float` or `dict`.

src/lenskit/metrics/_base.py
```diff
@@ -95,6 +145,11 @@ def measure_run(self, output: ItemListCollection, test: ItemListCollection, /) -
 
 class DecomposedMetric(Metric):
     """
+    Deprecated base class for decomposed metrics.
+    .. deprecated:: 2025.4
+        This class is deprecated and its functionality has been moved to :class:`Metric`.
+        It is scheduled for removal in 2026.
```
mdekstrand, line 189: I think we need a blank line before and after `.. deprecated`.

sushobhan2024: I checked another example `bulk.py` in the codebase and it has a blank line only before `.. deprecated`. So, I have now updated similar to that.

src/lenskit/metrics/_base.py
sushobhan2024: I have updated with the return data and included `float` and `dict`. What other data types should we allow for `measure_list` return?

src/lenskit/metrics/_base.py
sushobhan2024: I have now removed the docstring from the method implementation, and added a note in the class's docstring.

src/lenskit/metrics/_accum.py
```diff
@@ -0,0 +1,77 @@
...
+        Args:
+            metric: An instance of a Metric subclass.
+        """
+        self._metrics.append(metric)
+
+    def measure_list(
+        self,
+        output: ItemList,
+        test: ItemList,
+        user_id: Optional[Any] = None,
```
mdekstrand: This should be a kwargs of keys — not all evaluations have user IDs, and some may have more than one key. The keys should work like `ItemListCollection` (in fact, the same tuple type logic can be used).

src/lenskit/metrics/_accum.py
```diff
@@ -0,0 +1,77 @@
...
+        Measure and accumulate metrics for a single recommendation list.
+
+        Args:
+            output: The recommendation items for a user.
+            test: The ground-truth items for the same user.
+            user_id: Optional identifier for the user.
+        """
+        record = {"user_id": user_id}
+        for metric in self._metrics:
+            result = metric.measure_list(output, test)
```
mdekstrand: We need to keep the results around too — the aggregate will need them.

src/lenskit/metrics/_accum.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        Compute overall summary statistics by aggregating per-list metrics.
+
+        Returns:
+            A dictionary mapping metric labels to their aggregated summary values.
+        """
+        summaries = {}
+        df = self.list_metrics()
+        for metric in self._metrics:
+            values = df[metric.label].dropna().tolist()
```
mdekstrand: I don't think we should filter data here — if we are to drop NAs, the summarizer should do it. I also don't think we want to convert to `list` — those are inefficient (this is where we may need a little bit of a rethink on the types on the metric methods).

src/lenskit/metrics/_accum.py
```diff
@@ -0,0 +1,77 @@
...
+
+        Args:
+            output: The recommendation items for a user.
+            test: The ground-truth items for the same user.
+            user_id: Optional identifier for the user.
+        """
+        record = {"user_id": user_id}
+        for metric in self._metrics:
+            result = metric.measure_list(output, test)
+            record[metric.label] = metric.extract_list_metrics(result)
```
mdekstrand: Is this what we want if the list metrics return a dict? Possibly, but need to take a think about it.

src/lenskit/metrics/_accum.py
mdekstrand: Also, the summarizer should get the list measurement results, not the extracted metrics.

For most of metrics currently in LensKit, there is no difference. However, in the future, there will be: the intermediate results may not be metrics. This is also the case for RMSE: the per-list result for RMSE is the sum of squares and the number of ratings. Extracting a list metric converts those into an RMSE; computing the global micro-averaged RMSE sums them up to compute a global RMSE.

src/lenskit/metrics/_accum.py
```diff
@@ -0,0 +1,77 @@
...
+from typing import Any, Dict, List, Optional, Union
+
+import pandas as pd
+
+from lenskit.data import ItemList
+
+from ._base import Metric
+
+
+class MetricAccumulator:
```
mdekstrand, line 78: We need tests for this class.

src/lenskit/metrics/_accum.py
```diff
@@ -0,0 +1,77 @@
...
+            and columns correspnd to metrics.
+        """
+        return pd.DataFrame(self._results)
+
+    def summary_metrics(self) -> Dict[str, float]:
+        """
+        Compute overall summary statistics by aggregating per-list metrics.
+
+        Returns:
+            A dictionary mapping metric labels to their aggregated summary values.
```
mdekstrand: What if a metric produces more than one summary value?

mdekstrand (changes requested):
The API for `MetricAccumulator` looks pretty good. I was pondering whether it should return an analysis result object like `RecListAnalysis` currently does; I am not sure. What do you think? We will need to create analysis results for compatibility when we switch `RecListAnalysis` to use this code.

We do need some revision on the list ID API (it can't just be a user ID), and on how we're handling metric results (both the intermediate results from `measure_list`, and the extracted and summarized metrics). Two key points there:

- We need to save intermediate results, because summarization works on those, not on the extracted metrics.
- Both `extract_list_metrics` and `summarize` can return more than one metric value, and we need to make sure that we handle these the way we want to, and document how they are handled. Nested dictionaries can work, but quickly get annoying and take a lot of space. We might want instead to use dot-separated names: if a metric "RecipRank" produces a summary named "mean", that output value becomes "RecipRank.mean".

sushobhan2024:
@mdekstrand Yes, I think returning an object like `RunAnalysisResults` from the accumulator provides a consistent way to access both per-list and summary metrics. In this update, I have:

1. added `to_result` method that returns `RunAnalysisResults` object. 
2. added `normalize_metrics` method to flatten dictionary in case of more than one summary values. `extract_list_metrics` and `summarize` use this method.
3. refactored `RunAnalysis.measure` to used `MetricAccumulator`, and
4. added tests for the accumulator

I was thinking if intermediate results are needed anywhere else other than in summarization. Do we need to add a method to return intermediate results as well?

Also, should the summary include only computed metrics, or also registered but unmeasured ones? After adding a test for unmeasured registered metrics, I added `defaults` to the `to_result` method so they are included in the summary.

src/lenskit/metrics/_accum.py
```diff
@@ -0,0 +1,124 @@
...
+            summary.setdefault(key, value)
+
+        return RunAnalysisResult(
+            self.list_metrics(),
+            pd.Series(summary),
+            defaults,
+        )
+
+
+def normalize_metrics(label: str, value: float | dict[str, float] | None) -> dict[str, float]:
```
mdekstrand: Maybe a clearer name, since the are so many different things `normalize_metrics` could mean? Something like `metric_dict`?

src/lenskit/metrics/_base.py
mdekstrand: It can be anything. Vectors will be common, as will tuples (i.e. RMSE can return a tuple).

src/lenskit/metrics/_accum.py
```diff
@@ -0,0 +1,124 @@
...
+            A dictionary mapping metric labels to their aggregated summary values.
+        """
+        summaries = {}
+        for metric in self._metrics:
+            label = metric.label
+            values = [r[label] for r in self._intermediates if label in r]
+            if not values:
+                _log.warning("No intermediate values for metric '%s'; skipping summary.", label)
+                continue
+            result = metric.summarize(values)
```
mdekstrand: We should make this more efficient at some point but this is probably good enough for now.

mdekstrand (commented):
Need to get tests passing, but the design looks good to me now.

src/lenskit/metrics/_accum.py
sushobhan2024: I agree `normalize_metrics` is a bit generic. I will rename it to `to_metric_dict`.

mdekstrand:
@sushobhan2024 yes it should.

src/lenskit/metrics/_accum.py
```diff
@@ -0,0 +1,301 @@
...
+        """Validate that the accumulator is properly configured."""
+        seen = set()
+        for m in self.metrics:
+            lbl = m.label
+            if lbl in seen:
+                raise RuntimeError(f"duplicate metric: {lbl}")
+            seen.add(lbl)
+
+
+def to_result(acc: MetricAccumulator):
```
mdekstrand: Is this intended to be the usual way of getting metric results, or is it a compatibility shim?

If it is a compatibility shim, we should mark it as deprecated.

In any case, is there a reason not to make it a method of `MetricAccumulator`?

src/lenskit/metrics/_base.py
```diff
@@ -22,27 +25,79 @@ class Metric(ABC):
     Base class for LensKit metrics.  Individual metrics need to implement a
     sub-interface, such as :class:`ListMetric` and/or :class:`GlobalMetric`.
 
-    For simplicity in the analysis code, you cannot simply implement the
-    properties of this class on an arbitrary class in order to implement a
-    metric with all available behavior such as labeling and defaults; you must
-    actually extend this class.  This requirement may be relaxed in the future.
```
mdekstrand, line 28: This note seems important to keep, maybe as a `.. note::` towards the end of the class documentation.

src/lenskit/metrics/_base.py
```diff
@@ -22,27 +25,79 @@ class Metric(ABC):
...
+            - None if no per-list metrics are available
+        """
+        if isinstance(data, (float, int, np.floating, np.integer)):
+            return float(data)
+        elif isinstance(data, dict):
+            return data
+        else:
+            return None
+
+    def summarize(self, values: list[Any] | pa.Array | pa.ChunkedArray, /) -> dict[str, float]:
```
mdekstrand, line 86: Need documentation for this method.

src/lenskit/metrics/_base.py
```diff
@@ -22,27 +25,79 @@ class Metric(ABC):
...
+        if isinstance(values, (pa.Array, pa.ChunkedArray)):
+            values = values.to_pylist()
+
+        numeric_values = []
+        for v in values:
+            if v is not None:
+                if isinstance(v, (float, int, np.floating, np.integer)):
+                    numeric_values.append(float(v))
+                elif isinstance(v, dict) and "mean" in v:
+                    numeric_values.append(float(v["mean"]))
```
mdekstrand: Why are individual list measurements returning something named `mean`, and is this the correct handling? Shouldn't we just accumulate all of the values and take means of whatever metrics implementation computed?

src/lenskit/metrics/_base.py
```diff
@@ -72,6 +125,31 @@ def measure_list(self, output: ItemList, test: ItemList, /) -> float:
         """
         raise NotImplementedError()
 
+    def extract_list_metrics(self, data: Any, /) -> float | None:
+        """
+        Return the given per-list metric result.
+        """
+        if isinstance(data, (float, int, np.floating, np.integer)):
+            return float(data)
```
mdekstrand: if the data is a dict, maybe we should return that too? To make multi-metric list metrics easy.

src/lenskit/metrics/_base.py
mdekstrand: Since this summarization is only correct for `ListMetric` metrics, maybe leave it abstract here and provide this default implementation just in the `ListMetric` class. In any case, we shouldn't repeat this implementation in two places.

If we leave it abstract here, mark with `@abstractmethod`.

src/lenskit/metrics/_base.py
```diff
@@ -72,6 +125,31 @@ def measure_list(self, output: ItemList, test: ItemList, /) -> float:
...
+        if isinstance(values, (pa.Array, pa.ChunkedArray)):
+            values = values.to_pylist()
+
+        numeric_values = []
+        for v in values:
+            if v is not None:
+                if isinstance(v, (float, int, np.floating, np.integer)):
+                    numeric_values.append(float(v))
+                elif isinstance(v, dict) and "mean" in v:
+                    numeric_values.append(float(v["mean"]))
```
mdekstrand: We should make this more efficient (vectorized) but we can do that later.

src/lenskit/metrics/_base.py
```diff
@@ -22,27 +25,79 @@ class Metric(ABC):
...
         The metric's default label in output.
-
         The base implementation returns the class name by default.
         """
         return self.__class__.__name__
 
     def __str__(self):
         return f"Metric {self.label}"
 
+    def measure_list(self, output: ItemList, test: ItemList, /) -> Any:
```
mdekstrand, line 63: Mark this with `@abstractmethod` to signal that it is an abstract method.

src/lenskit/metrics/bulk.py
```diff
@@ -238,68 +178,37 @@ def measure(
...
-        self._validate_setup()
-        index = pd.MultiIndex.from_tuples(outputs.keys())
-        index.names = list(outputs.key_fields)
 
-        lms = [m for m in self.metrics if m.is_listwise or m.is_decomposed]
-        gms = [m for m in self.metrics if m.is_global]
-        list_results = pd.DataFrame({m.label: np.nan for m in lms}, index=index)
-        list_intermediates = {m.label: [] for m in self.metrics if m.is_decomposed}
+        This method now uses the MetricAccumulator internally, which provides
+        better separation of concerns and supports external evaluation loops.
```
mdekstrand: Is this note in the documentation string important to understanding the interface of the method?

src/lenskit/metrics/bulk.py
```diff
@@ -238,68 +178,37 @@ def measure(
...
-                m.label: (
-                    m.metric.global_aggregate(list_intermediates[m.label])
-                    if m.is_decomposed
-                    else m.measure_run(outputs, test)
-                )
-                for m in self.metrics
-                if m.is_global or m.is_decomposed
-            },
-            dtype=np.float64,
-        )
```
mdekstrand, line 295: With removing this code, global metrics support seems to be fully removed. Is it happening somewhere else that I am missing?

src/lenskit/metrics/_accum.py
```diff
@@ -0,0 +1,301 @@
...
+            try:
+                if wrapper.is_decomposed:
+                    data = wrapper.compute_list_data(output, test)
+                    self._list_data[wrapper.label].append(data)
+                    list_metric = wrapper.extract_list_metrics(data)
+                    self._list_metrics[wrapper.label].append(list_metric)
+                else:
+                    metric_value = wrapper.measure_list(output, test)
+                    self._list_data[wrapper.label].append(metric_value)
+                    self._list_metrics[wrapper.label].append(metric_value)
```
mdekstrand: We shouldn't need this complexity any more, should we? The advantage of making everything (except global metrics) a `Metric` is that we can just call `Metric` methods.

mdekstrand: We will still need something for global metrics.

src/lenskit/metrics/_accum.py
```diff
@@ -0,0 +1,301 @@
...
+        else:
+            index = pd.Index(self._list_keys)
+
+        data = {}
+        for wrapper in self.metrics:
+            metric_results = self._list_metrics[wrapper.label]
+
+            if metric_results and isinstance(metric_results[0], dict):
+                for result in metric_results:
+                    if isinstance(result, dict):
```
mdekstrand: this is going to cause mixed-type results (which should never happen, but aren't impossible) to behave incorrectly.

mdekstrand: case where it may happen: if the metric returns `None` instead of a dict for an empty / unmeasurable list, then that row will be fully skipped in this result.

src/lenskit/metrics/_accum.py
```diff
@@ -0,0 +1,301 @@
...
+            else:
+                wl = m.__name__  # type: ignore
+        else:
+            wl = label
+
+        if default is None:
+            if isinstance(m, ListMetric):
+                default = m.default
+            else:
+                default = 0.0
```
mdekstrand: What is the basis for this being correct?

src/lenskit/metrics/bulk.py
```diff
@@ -220,7 +150,17 @@ def add_metric(
...
-        self.metrics.append(_wrap_metric(metric, label, default))
+        self._accumulator.add_metric(metric, label, default)
+
+    @property
+    def metrics(self) -> list:
+        """
+        The list of metrics to compute.
+
+        .. deprecated:: 2025.4
+            Access metrics through the accumulator interface instead.
```
mdekstrand, line 164: If we want the user to get the metrics through the accumulator, then we should expose the accumulator. Maybe just make it a public attribute.

mdekstrand:
I suspect the evaluation-based test is a change in behavior in the RMSE metric.

src/lenskit/metrics/_accum.py
sushobhan2024: I think I copy-pasted this from the old codebase. Are you asking me to change `ListMetric` to `Metric` in the `isinstance` check, or are you suggesting I should write a test case to verify this logic?

src/lenskit/metrics/_accum.py
sushobhan2024: Should I process each result individually instead of making type assumptions based on first result?

src/lenskit/metrics/_accum.py
sushobhan2024: You're right. I have now simplified this. For the global metrics, should we keep the logic in RunAnalysis as in the old code, or add a method in the accumulator?

src/lenskit/metrics/_accum.py
mdekstrand: My question is specifically about defaulting the default to 0. There are many cases in which that would not be correct.

Dealing with defaults is complex and subtle, and may actually be more complexity than is good to have. With just a list metric, it was easy, but with a more complex notion of metric outputs, it's less straightforward. However, metrics are also now responsible for their own summarization, which helps us — the primary purpose of the default was to correctly input something into the summarization process.

Therefore, we might actually want to pull the use of the `default` logic into the `ListMetric` implementation of `summarize`, and make a note for LensKit 2026 to make metrics directly return their default values when appropriate. We have a period we need to keep some compatibility, but we might want to yeet this concept of “defaults” as a metric property.

src/lenskit/metrics/bulk.py
sushobhan2024: I realize I have been assigning this to `summarize` instead which is incorrect.

src/lenskit/metrics/_base.py
```diff
@@ -22,27 +25,79 @@ class Metric(ABC):
...
+        numeric_values = []
+        for v in values:
+            if v is not None:
+                if isinstance(v, (float, int, np.floating, np.integer)):
+                    numeric_values.append(float(v))
+                elif isinstance(v, dict) and "mean" in v:
+                    numeric_values.append(float(v["mean"]))
+
+        if not numeric_values:
+            return {"mean": 0.0}
```
mdekstrand: We should return `None` for empty results.

src/lenskit/metrics/_base.py
```diff
@@ -103,27 +193,44 @@ class DecomposedMetric(Metric):
...
+    def extract_list_metric(self, data: Any, /) -> float | None:
         """
         Extract a single-list metric from the per-list measurement result (if
         applicable).
 
         Returns:
             The per-list metric, or ``None`` if this metric does not compute
             per-list metrics.
+
+        Use `extract_list_metrics` in `Metric` for new implementations.
```
mdekstrand: reStructuredText uses double-backticks for monospace/code, not the single backticks of Markdown, so this would be:

```rst
Use ``extract_list_metrics`` in ``Metric`` for new implementations.
```

However, since we are referencing a method, it is better to use a method reference:

```rst
Implement :meth:`Metric.extract_list_metrics` in new implementations.
```

src/lenskit/metrics/bulk.py
```diff
@@ -238,68 +181,69 @@ def measure(
...
+                        summary = metric_wrapper.summarize(clean_data)
+                        if isinstance(summary, dict):
+                            result = summary.get("mean", 0.0)
+                        else:
+                            result = float(summary)
+                    else:
+                        result = metric_wrapper.default or 0.0
+                else:
+                    _log.warning("unexpected metric type for %s", metric_wrapper.label)
+                    result = 0.0
```
mdekstrand: We shouldn't have these cases anymore — the thing the global metric logic needs to handle is truly global metrics, decomposed metrics should be handled entirely by the accumulator.

src/lenskit/metrics/_accum.py
```diff
@@ -38,8 +38,10 @@ class MetricWrapper:
 
     @property
     def is_listwise(self) -> bool:
-        "Check if this metric is listwise."
-        return isinstance(self.metric, (ListMetric, Callable))
+        "Check if this metric is listwise (includes ListMetric and callable functions)."
+        return isinstance(self.metric, (ListMetric, Callable)) or (
+            isinstance(self.metric, Metric) and hasattr(self.metric, "measure_list")
```
mdekstrand: we shouldn't need both of these tests — `measure_list` is defined by `Metric`.

mdekstrand: Maybe this whole test should just test that it does not implement `GlobalMetric`? (I don't remember the exact relationship between `GlobalMetric` and `DecomposedMetric` right now.)

sushobhan2024: The accumulator code treats anything that is not `GlobalMetric` as list wise. Should we can use a negative test, something like `not isinstance(self.metric, GlobalMetric)`, or use the code as in the old codebase? I have reverted to the old code now.

There were 2 eval tests failing after I pushed the code. However, running the tests in my computer skipped those tests. I am investigating why these are failing.

sushobhan2024: Those tests are being skipped locally because ML100k data is not available.

mdekstrand:
@sushobhan2024 This is the important bit:
> ```
> error: rustc 1.86.0 is not supported by the following package:
>    nalgebra@0.34.0 requires rustc 1.87.0
> ```

Run `rustup update` to get a newer Rust.

tests/eval/test_metric_accumulator.py
```diff
@@ -145,15 +145,46 @@ def summarize(self, values):
...
+    acc.add_metric(CallableMetric(), "callable")
+    il = ItemList([1, 2])
+    test_il = ItemList([1])
+    acc.measure_list(il, test_il, user="u1")
+    acc.measure_list(il, test_il, user="u2")
+    summary_df = acc.summary_metrics()
+
+    assert summary_df.loc["callable", "mean"] == approx(12)
+    assert pd.isna(summary_df.loc["callable", "median"])
+    assert pd.isna(summary_df.loc["callable", "std"])
```
mdekstrand: Since we are testing several different things, it would be better to separate these into separate tests, one for each specific case / problem being tested.

src/lenskit/metrics/_accum.py
```diff
@@ -0,0 +1,295 @@
...
+    def measure_list(self, list: ItemList, test: ItemList) -> Any:
+        """Get intermediate measurement data from the metric."""
+        if isinstance(self.metric, Callable):
+            return self.metric(list, test)
+        return self.metric.measure_list(list, test)
+
+    def extract_list_metrics(self, intermediate_data: Any) -> float | dict[str, float] | None:
+        """Extract per-list metrics from intermediate data."""
+        result = self.metric.extract_list_metrics(intermediate_data)
+        return intermediate_data if result is None else result
```
mdekstrand: This should just return `result` — if the metric doesn't return a value from `extract_list_metrics`, then it doesn't support list metrics. We can't assume that the intermediate data is list metrics.

src/lenskit/metrics/_accum.py
```diff
@@ -0,0 +1,295 @@
...
+
+        if not numeric_values:
+            return {"mean": None, "median": None, "std": None}
+
+        arr = np.array(numeric_values, dtype=np.float64)
+        return {
+            "mean": float(np.mean(arr)),
+            "median": float(np.median(arr)),
+            "std": float(np.std(arr, ddof=1)) if len(arr) > 1 else 0.0,
+        }
```
mdekstrand: I don't think we need default summarization — `ListMetric` implements default summarization logic, and for other metrics we can't assume that it is meaningful to compute summary statistics over intermediate data. If we did provide default summarization, it would need to be over the extracted list metrics, and I think we should defer that to a future enhancement if it's even something we want.

src/lenskit/metrics/_accum.py
```diff
@@ -0,0 +1,295 @@
...
+        return intermediate_data if result is None else result
+
+    def summarize(self, values: list[Any] | pa.Array | pa.ChunkedArray) -> dict[str, float | None]:
+        """Aggregate intermediate values into summary statistics."""
+        result = self.metric.summarize(values)
+        if result is None:
+            return self._default_summarize(values)
+        if isinstance(result, dict):
+            return result
+        return {"mean": float(result), "median": None, "std": None}
```
mdekstrand: We can't assume that a single result is a mean — it could just be a single summary value. For example, Gini will return a single value.

src/lenskit/metrics/_accum.py
mdekstrand: This still seems open — a default of `0.0` doesn't seem appropriate, unless there's something I'm missing.

src/lenskit/metrics/_base.py
```diff
@@ -22,27 +25,76 @@ class Metric(ABC):
...
+        metric with all available behavior such as labeling and defaults;
+        you must actually extend this class. This requirement may be relaxed
+        in the future.
+    """
+
+    default: ClassVar[float | None] = 0.0
+    """
+    The default value to infer when computing statistics over missing values.
+    If ``None``, no inference is done (necessary for metrics like RMSE, where
+    the missing value is theoretically infinite).
```
mdekstrand: "infer" -> "impute", "inference" -> "imputation".

mdekstrand: Should `default` live on `ListMetric` instead of `Metric`? A single `default` value for a metric, that may compute more than one metric, doesn't seem to make sense.

src/lenskit/metrics/ranking/_gini.py
```diff
@@ -50,24 +50,11 @@ class ListGini(GiniBase):
...
-        k:
-            The maximum recommendation list length.
-        items:
-            The total number of items, a data frame or series of item data, or a
-            dataset. If a frame or series is provided, its length will be used
-            as the number of items.  If a dataset is provided, its item count
-            will be used.
-
-    Stability:
-        Caller
```
mdekstrand: Why remove this?

mdekstrand (changes requested):
Almost there! We just need to drop some excess logic that assumes more about metrics than the interface guarantees, and I think we'll be ready to merge.

src/lenskit/metrics/ranking/_gini.py
sushobhan2024: Will keep this. This should be a mistake. I don't remember removing.

PR #803, opened 2025-06-21
added binarize_ratings function in DatasetBuilder

vibhor-5:
This PR adds a binarize_ratings method to the DatasetBuilder class, enabling users to easily binarize the ratings in a relationship class (typically "rating"). This is useful for preparing implicit feedback datasets or filtering out low ratings.

Details

DatasetBuilder.binarize_ratings(cls="rating", min_rating=1.0, method="remove")

cls: The relationship class to binarize (default: "rating").
min_rating: Minimum value to consider as positive.
method:
- "zero": Set ratings to 1 if >= min_rating, else 0.
- "remove": Remove rows where rating < min_rating (default).

### DIFF
--- a/src/lenskit/data/builder.py
+++ b/src/lenskit/data/builder.py
@@ -587,6 +587,42 @@ class DatasetBuilder:
 
         self._tables[cls] = tbl
 
+        def binarize_ratings(
+            self,
+            cls: str = "rating",
+            min_rating: float = 1.0,
+            method: str = "remove",
+        ):
+            """
+            Binarize the ratings in a relationship class.
+
+            Args:
+                cls (str): The relationship class to binarize (default: "rating").
+                min_rating (float): Minimum rating to consider as positive.
+                method (str): 'zero' to set ratings to 0/1, 'remove' to drop rows below min_rating.
+            """
+            tbl = self._tables.get(cls, None)
+            if tbl is None:
+                raise ValueError(f"relationship class {cls} is empty")
+
+            if "rating" not in tbl.column_names:
+                raise ValueError(f"relationship class {cls} does not have a 'rating' column")
+
+            min_rating: pa.scalar = pa.scalar(min_rating, tbl.column("rating").type)
+            if method == "zero":
+                # Set rating to 1 if >= min_rating, else 0
+                mask = pc.greater_equal(tbl.column("rating"), min_rating)
+                binarized = pc.cast(mask, pa.int32())
+                tbl = tbl.set_column(tbl.schema.get_field_index("rating"), "rating", binarized)
+            elif method == "remove":
+                # Keep only rows where rating >= min_rating
+                mask = pc.greater_equal(tbl.column("rating"), min_rating)
+                tbl = tbl.filter(mask)
+            else:
+                raise ValueError("method must be 'zero' or 'remove'")
+
+            self._tables[cls] = tbl
+
     def clear_relationships(self, cls: str):
         """
         Remove all records for a specified relationship class.


src/lenskit/data/builder.py
```diff
@@ -587,6 +587,42 @@ def filter_interactions(
...
+                raise ValueError(f"relationship class {cls} is empty")
+
+            if "rating" not in tbl.column_names:
+                raise ValueError(f"relationship class {cls} does not have a 'rating' column")
+
+            min_rating: pa.scalar = pa.scalar(min_rating, tbl.column("rating").type)
+            if method == "zero":
+                # Set rating to 1 if >= min_rating, else 0
+                mask = pc.greater_equal(tbl.column("rating"), min_rating)
+                binarized = pc.cast(mask, pa.int32())
```
mdekstrand: Let's make this a `float32`, ratings are more commonly used as floats.

src/lenskit/data/builder.py
```diff
@@ -587,6 +587,42 @@ def filter_interactions(
 
         self._tables[cls] = tbl
 
+        def binarize_ratings(
+            self,
+            cls: str = "rating",
+            min_rating: float = 1.0,
```
mdekstrand: Let's make this default 3.0.

src/lenskit/data/builder.py
```diff
@@ -587,6 +587,42 @@ def filter_interactions(
...
+            method: str = "remove",
+        ):
+            """
+            Binarize the ratings in a relationship class.
+
+            Args:
+                cls (str): The relationship class to binarize (default: "rating").
+                min_rating (float): Minimum rating to consider as positive.
+                method (str): 'zero' to set ratings to 0/1, 'remove' to drop rows below min_rating.
```
mdekstrand: With type annotations in the declaration, we don't use them in the documentation string as well.

src/lenskit/data/builder.py
mdekstrand: Maybe call it `min_pos_rating`, just for clarity?

src/lenskit/data/builder.py
vibhor-5: changed the data type to `pa.float32`

vibhor-5:
made all the minor changes ,
about the test, should i write a new file for the tests or write it in the `test_builder_filter` file itself?

tests/data/test_builder_binarize.py
```diff
@@ -0,0 +1,53 @@
...
+
+from pytest import approx, mark, raises, skip, warns
+
+from lenskit.data import Dataset, DatasetBuilder
+
+
+def test_binarize_zero(ml_ds: pd.DataFrame) -> None:
+    """Test that binarize removes items."""
+    dsb = DatasetBuilder()
+    dsb.add_interactions("rating", ml_ds, entities=["user", "item"], missing="insert", default=True)
```
mdekstrand: I think this is likely the problem: `ml_ds` is a `Dataset`, not a data frame.

If you set up the builder as a copy with `dsb = DatasetBuilder(ml_ds)`, and remove this line, then it should get past this problem with the tests.

vibhor-5: i tried that, and i tried even using 
```
ratings_df = ml_ds.interactions().pandas()
dsb = DatasetBuilder()
 dsb.add_interactions(
        "rating", ratings_df, entities=["user", "item"], missing="insert", default=True
    )
```
as well but its still not working

showing error 
```
_____________________ ERROR at setup of test_binarize_zero _____________________

    @pytest.fixture(scope="session")
    def ml_ds_unchecked() -> Generator[Dataset, None, None]:
        """
        Fixture to load the MovieLens dataset, without checking for modifications.
    
        Usually use :func:`ml_ds` instead.
        """
>       yield load_movielens(ml_test_dir)

src/lenskit/testing/_movielens.py:67: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/lenskit/data/movielens.py:453: in load_movielens
    return ml.dataset()
src/lenskit/data/movielens.py:372: in dataset
    return dsb.build()
src/lenskit/data/builder.py:979: in build
    return Dataset(self.build_container())
src/lenskit/data/dataset.py:90: in __init__
    self._init_caches()
src/lenskit/data/dataset.py:165: in _init_caches
    self._relationships[name] = MatrixRelationshipSet(self, name, schema, tbl)
src/lenskit/data/relationships.py:211: in __init__
    self._init_structures(ds_name=ds.name)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <lenskit.data.relationships.MatrixRelationshipSet object at 0x1110ed0d0>

    def _init_structures(self, *, ds_name: str | None = None, _trust_table_sort: bool = False):
        log = _log.bind(dataset=ds_name, relationship=self.name)
    
        # order the table to compute the sparse matrix
        log.debug("setting up entity information")
        entities = list(self.schema.entities.keys())
        row, col = entities
    
        self.row_type = row
        self.col_type = col
    
        e_cols = [num_col_name(e) for e in entities]
        log.debug("checking relationship table sorting")
        if _trust_table_sort or _data_accel.is_sorted_coo(self._table.to_batches(), *e_cols):
            log.debug("relationship table already sorted 😊")
        else:
            log.warning("sorting relationship table (might take time)")
            self._table = self._table.sort_by([(c, "ascending") for c in e_cols])
    
        self._table = self._table.combine_chunks()
    
        # compute the row pointers
        log.debug("computing CSR data")
        n_rows = len(self.row_vocabulary)
        row_sizes = np.zeros(n_rows + 1, dtype=np.int64)
        self._row_nums = self._table.column(e_cols[0]).combine_chunks()  # type: ignore
        rsz_struct = pc.value_counts(self._row_nums)
        rsz_nums = rsz_struct.field("values")
        rsz_counts = rsz_struct.field("counts").cast(pa.int32())
        row_sizes[np.asarray(rsz_nums) + 1] = rsz_counts
        self._row_ptrs = np.cumsum(row_sizes, dtype=np.int64)  # type: ignore
    
        self._col_nums = self._table.column(e_cols[1]).combine_chunks()  # type: ignore
        self._structure = SparseRowArray.from_arrays(
            self._row_ptrs,
            self._col_nums,
            shape=(len(self.row_vocabulary), len(self.col_vocabulary)),
        )
    
        # make the index
        log.debug("computing row-column index")
>       self._rc_set = RowColumnSet(self._structure)
E       TypeError: invalid array type LargeList(Field { name: "item", data_type: Int32, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {"ARROW:extension:name": "lenskit.sparse_index", "ARROW:extension:metadata": "{\"dimension\": 9125}"} }), expected List

src/lenskit/data/relationships.py:254: TypeError
```

mdekstrand: That's an interesting problem, and it looks like a bug in `RowColumnSet`. I'm not sure why it wasn't failing before.

Can you update your branch against the current `main` and see if that clears this up? I suspect it might (we just recently made some changes to large / non-large lists in dataset).

vibhor-5: Just tried that , still the same error
Idk why is the large list type being rejected 
Ill try to see in the data -relationship pipeline if im doing something wrong with the ml_ds argument maybe

mdekstrand: It's also possible we have a lingering bug that is shaken loose by this code path but wasn't in any of our previous tests paths.

mdekstrand: Oh, I think I see the problem, reading the stack trace again — this is in code that is completely untouched.

I think the Rust extension module is out-of-date. Run `uv sync` again, and if that doesn't rebuild the module, `invoke build`.

vibhor-5: YESS OMG thankss that workedd!!! i pushed the working code now with corrected tests

vibhor-5: This was my very first PR so apologies for some dumb mistakes . Thankyou for your support really appreciate it

vibhor-5: hey @mdekstrand any more changes needed ?

mdekstrand: @vibhor-5 Thanks! I'm on vacation right now, might not get to reviewing changes until late next week.

mdekstrand (approved):
Looks good to merge when we clear the test failure. #805 should have the fix for that, will merge to main after it passes CI.

PR #812, opened 2025-08-13
Add support for repeated interactions 

FroggoLight:
This adds support for allowing data with repeated and duplicated interactions to be built as a RelationshipSet and allowing it to be converted as a MatrixRelationshipSet


src/lenskit/data/builder.py
```diff
@@ -165,8 +169,9 @@ def add_relationship_class(
         self,
         name: str,
         entities: RelationshipEntities,
-        allow_repeats: bool = True,
+        repeats: bool = True,
```
mdekstrand: Then name and behavior of this option are part of a stable API, so we need to keep them the same.

src/lenskit/data/builder.py
```diff
@@ -165,8 +169,9 @@ def add_relationship_class(
...
         interaction: bool = False,
+        remove_duplicates: bool | Literal["exact"] = False,
```
mdekstrand: We went back and forth on a couple designs here, but where we ultimately landed is that whether repeats are _allowed_ is a property of the relationship class that gets stored in the schema, but whether they are _removed_ is a property of adding relationships. Therefore, `remove_duplicates` should be an option to `add_relationships` and `add_interactions`, but _not_ `add_relationship_class` (or `RelationshipSchema`).

src/lenskit/data/builder.py
```diff
@@ -191,13 +196,13 @@ def add_relationship_class(
                 are allowed.
             interaction:
                 Whether this is an interaction relationship.
+            remove_duplicates:
+                If ``True``, exact duplicate interactions will be removed.
```
mdekstrand: See above, this won't be on the relationship class.

src/lenskit/data/builder.py
```diff
@@ -327,9 +333,10 @@ def add_relationships(
         *,
         entities: RelationshipEntities | None = None,
         missing: MissingEntityAction = "error",
-        allow_repeats: bool = True,
+        repeats: bool = True,
```
mdekstrand: See above, can't change this option name.

src/lenskit/data/builder.py
```diff
@@ -359,6 +366,8 @@ def add_relationships(
                 Whether this is an interaction relationship or not; can be
                 ``"default"`` to indicate this is the default interaction
                 relationship.
+            remove_duplicates:
+                If ``True``, exact duplicate interactions will be removed.
```
mdekstrand: Need to update this to more clearly reflect the full (intended) behavior, and probably also change the name to be clearer. I think it would be better to call it `remove_repeats`, and it has 3 options (as currently allowed by the type):

- `False`: repeats are not removed
- `True`: repeats are removed
- `"exact"`: only exact duplicates are removed

src/lenskit/data/builder.py
```diff
@@ -439,13 +449,7 @@ def add_relationships(
...
             dupes = ndf.duplicated()
             if np.any(dupes):
-                if rc_def.repeats.is_allowed:
-                    rc_def.repeats = AllowableTroolean.PRESENT
-                else:
-                    raise DataError(
-                        f"repeated interactions not allowed for relationship class {cls}"
-                    )
-
+                self._resolve_repeated_interactions(t=new_table, rel=rc_def)
```
mdekstrand: Let's write the new logic in a way that we don't need to first convert to Pandas and check for duplicates — it will save time. However, we do need to make sure that `ALLOWED` is properly upgraded to `PRESENT` when there are repeats.

src/lenskit/data/builder.py
```diff
@@ -458,8 +462,9 @@ def add_interactions(
         *,
         entities: RelationshipEntities | None = None,
         missing: MissingEntityAction = "error",
-        allow_repeats: bool = True,
+        repeats: bool = True,
         default: bool = False,
+        remove_duplicates: bool | Literal["exact"] = False,
     ) -> None:
```
mdekstrand: Same comments as above

src/lenskit/data/builder.py
```diff
@@ -489,20 +494,23 @@ def add_interactions(
...
                 be ``"error"`` or ``"insert"``.
-            allow_repeats:
-                Whether repeated interactions are allowed.
+            repeats:
+                Whether repeated interactions are allowed, forbidden, or removed.
             default:
                 If ``True``, set this as the default interaction class (if the
                 dataset has more than one interaction class).
+            remove_duplicates:
+                If ``True``, exact duplicate interactions will be removed.
```
mdekstrand: Same comments as above re: names and documentation.

src/lenskit/data/builder.py
```diff
@@ -938,11 +946,18 @@ def build_container(self) -> DataContainer:
...
+                        e_cols = [e + "_num" for e in rel.entities.keys()]
+                        if not _data_accel.is_sorted_coo(t.to_batches(), *e_cols):
+                            log.debug("sorting non-repeating relationship %s", n)
+                            t = t.sort_by([(c, "ascending") for c in e_cols])
+                    if rel.remove_duplicates and rel.remove_duplicates != "exact":
+                        t = self._remove_repeated_relationships(t, rel)
+                    if rel.remove_duplicates == "exact":
+                        temp_df = t.to_pandas()
+                        temp_df.drop_duplicates(inplace=True)
+                        t = pa.Table.from_pandas(temp_df, preserve_index=False)
```
mdekstrand: Let's put duplicate and repeat handling into the relationship-adding only.

src/lenskit/data/builder.py
```diff
@@ -960,6 +975,17 @@ def save(self, path: str | PathLike[str]):
...
+    def _remove_repeated_relationships(self, t: pa.Table, rel: RelationshipSchema) -> pa.Table:
+        if "timestamp" in t.column_names:
+            t = t.sort_by([("timestamp", "ascending")])
+        t_modified = t.add_column(0, "_row_number", pa.array(np.arange(t.num_rows)))
+        t_modified = t_modified.group_by(
+            [entity + "_num" for entity in rel.entity_class_names]
+        ).aggregate([("_row_number", "max")])
+        t_modified = t_modified.sort_by([("_row_number_max", "ascending")])
+        t = t.take(t_modified.column("_row_number_max"))
+        return t
```
mdekstrand, line 1101: This looks sound.

src/lenskit/data/builder.py
```diff
@@ -970,6 +996,21 @@ def _resolve_entity_ids(
...
+                temp_df = t.to_pandas()
+                temp_df.drop_duplicates(inplace=True)
+                t = pa.Table.from_pandas(temp_df, preserve_index=False)
+            else:
+                t = self._remove_repeated_relationships(t, rel)
+
+        if rel.repeats.is_allowed and not remove_repeat:
+            rel.repeats = AllowableTroolean.PRESENT
+        elif rel.repeats.is_forbidden and not remove_repeat:
+            raise DataError("repeated interactions not allowed for relationship class")
```
mdekstrand: This needs to return the updated table — assigning to `t` is not sufficient, as it does not return the data outside of this function.

Also consider the relationship with `_remove_repeated_relationships` and think about either consolidating, or having `_remove_repeated…` and `_remove_duplicate…` for consistent clarity.

src/lenskit/data/relationships.py
```diff
@@ -67,6 +66,8 @@ class RelationshipSet:
     The name of the relationship class for these relationships.
     """
     schema: RelationshipSchema
+    dataset: Dataset
```
mdekstrand: we stopped passing this in, so remove the field too.

src/lenskit/data/relationships.py
```diff
@@ -67,6 +66,8 @@ class RelationshipSet:
...
+    mat_rset: dict[tuple[str, str], MatrixRelationshipSet]
```
mdekstrand: This is only for internal usage, so prefix with `_` (i.e. name it `_matrix_sets`).

src/lenskit/data/relationships.py
```diff
@@ -78,18 +79,35 @@ class RelationshipSet:
 
     def __init__(
         self,
-        ds: Dataset,
         name: str,
+        vocab: dict[str, Vocabulary],
```
mdekstrand: Since this is more than one vocabulary, better to name it `vocabularies` (or `vocabs`, but we don't need to save space).

src/lenskit/data/relationships.py
```diff
@@ -78,18 +79,35 @@ class RelationshipSet:
...
+            "table": self._table,
+            "vocabularies": self._vocabularies,
+        }
+
+    def __setstate__(self, state):
+        self.name = state["name"]
+        self.schema = state["schema"]
+        self._link_cols = state["columns"]
+        self._table = state["table"]
+        self._vocabularies = state["vocabularies"]
```
mdekstrand, line 106: good!

src/lenskit/data/relationships.py
```diff
@@ -103,7 +121,10 @@ def attribute_names(self) -> list[str]:
 
     def count(self):
         if "count" in self._table.column_names:  # pragma: nocover
-            raise NotImplementedError()
+            count_column = self._table.column("count").combine_chunks().slice(0)
```
mdekstrand: We shouldn't need `combine_chunks` or `slice` here.

src/lenskit/data/relationships.py
```diff
@@ -103,7 +121,10 @@ def attribute_names(self) -> list[str]:
...
+            count_column.fill_null(pa.scalar(1, type=pa.int8()))
```
mdekstrand: If this doens't work without `combine_chunks`, use `pc.fill_null` instead. With `pc.fill_null`, possibly also `column.fill_null`, we can also just pass `1` without passing it through `pa.scalar` (the type checker doesn't like it but it works, another problem with the PyArrow type stubs).

src/lenskit/data/relationships.py
```diff
@@ -163,19 +184,86 @@ def pandas(self, *, attributes: str | list[str] | None = None, ids=False) -> pd.
...
     def matrix(
-        self, *, combine: MAT_AGG | Mapping[str, MAT_AGG] | None = None
+        self, *, row_entity: str = "user", col_entity: str = "item"
+    ) -> MatrixRelationshipSet:  # pragma: nocover
+        if (row_entity, col_entity) in self.mat_rset:
+            return self.mat_rset[(row_entity, col_entity)]
+        else:
+            new_mat_rset = self._make_matrix(row_entity=row_entity, col_entity=col_entity)
+            self.mat_rset[(row_entity, col_entity)] = new_mat_rset
+            return new_mat_rset
```
mdekstrand: It doesn't really matter here, but style tip that speeds this up in cases where we do it a lot:

```python
mat = self.mat_rset.get((row_entity, col_entity), None)
if mat is None:
    mat = self._make_matrix(…)
    self.mat_rset[(row_entity, col_entity)] = mat
return mat
```

This way, you only do 1 dictionary lookup when the data is present instead of 2. This is infrequent enough it isn't going to be a noticable speed difference, but in other cases it might be, so it's a useful trick to know.

src/lenskit/data/relationships.py
```diff
@@ -163,19 +184,86 @@ def pandas(self, *, attributes: str | list[str] | None = None, ids=False) -> pd.
...
 
         Args:
-            combine:
-                The method for combining attribute values for repeated
-                relationships or interactions.  Can either be a single strategy
-                or a mapping from attribute names to combination strategies.
+            row_entity:
+                The specified row entity of the matrix
+            col_entity:
+                The specified column entity of the matrix
```
mdekstrand, line 194: Good.

src/lenskit/data/relationships.py
```diff
@@ -163,19 +184,86 @@ def pandas(self, *, attributes: str | list[str] | None = None, ids=False) -> pd.
...
+            raise FieldError(self.name, row_entity)
+
+        if col_entity not in self.schema.entities.keys():
+            raise FieldError(self.name, col_entity)
+
+        if col_entity == row_entity:
+            raise ValueError("row and column entity should not be the same")
+
+        e_dict: dict[str, str | None]
+        e_dict = {row_entity: None, col_entity: None}
```
mdekstrand: There is a very subtle correctness bit here: these should map to `self.schema.entities[row_entity]` (resp. `col_entity`), instead of `None`, in order to be correct in the case where the relationship type has mapped entity names. I don't think we have any tests for this right now, and I'm not sure it's even possible to build such a relationship set yet, but it's supported by the schema. At the very least, leave a `FIXME:` comment here to flag that it needs to change when mapped entity names are supported.

src/lenskit/data/relationships.py
```diff
@@ -163,19 +184,86 @@ def pandas(self, *, attributes: str | list[str] | None = None, ids=False) -> pd.
...
+            entities=e_dict,
+            interaction=self.schema.interaction,
+            repeats=self.schema.repeats,
+            attributes=self.schema.attributes,
+            remove_duplicates=self.schema.remove_duplicates,
+        )
+        new_table = self._table
+
+        if "timestamp" in new_table.column_names:
+            new_table = new_table.sort_by([("timestamp", "ascending")])
```
mdekstrand: decision point: do we always sort tables when we build the dataset? If we do, then we can save the sort here, and speed up this method — sorting is slow.

src/lenskit/data/relationships.py
```diff
@@ -163,19 +184,86 @@ def pandas(self, *, attributes: str | list[str] | None = None, ids=False) -> pd.
...
+
+        if "count" not in new_table.column_names:
+            new_table = new_table.add_column(
+                new_table.num_columns, "count", pa.array(np.ones(new_table.num_rows))
+            )
```
mdekstrand: For the case where we do not have a `count` column, it would be better to use a `count` aggregate, instead of creating a column of ones and summing them. It will probably be faster, and will definitely require less memory.

src/lenskit/data/relationships.py
```diff
@@ -202,13 +290,13 @@ class MatrixRelationshipSet(RelationshipSet):
 
     def __init__(
         self,
-        ds: Dataset,
         name: str,
+        vocab: dict[str, Vocabulary],
```
mdekstrand: Same comment above re: the name of this

src/lenskit/data/relationships.py
```diff
@@ -163,19 +184,86 @@ def pandas(self, *, attributes: str | list[str] | None = None, ids=False) -> pd.
...
+            return new_mat_rset
+
+    def _make_matrix(
+        self, *, row_entity: str = "user", col_entity: str = "item"
     ) -> MatrixRelationshipSet:  # pragma: nocover
         """
         Convert this relationship set into a matrix, coalescing duplicate
         observations.
 
         Args:
```
mdekstrand, line 190: Move the documentation to the `matrix` method, so it's on the public one.

src/lenskit/data/relationships.py
```diff
@@ -163,19 +184,86 @@ def pandas(self, *, attributes: str | list[str] | None = None, ids=False) -> pd.
...
+        if (row_entity, col_entity) in self.mat_rset:
+            return self.mat_rset[(row_entity, col_entity)]
+        else:
+            new_mat_rset = self._make_matrix(row_entity=row_entity, col_entity=col_entity)
+            self.mat_rset[(row_entity, col_entity)] = new_mat_rset
+            return new_mat_rset
+
+    def _make_matrix(
+        self, *, row_entity: str = "user", col_entity: str = "item"
     ) -> MatrixRelationshipSet:  # pragma: nocover
```
mdekstrand: We're actually writing this now, and should be testing it, so remove `# pragma: nocover`.

src/lenskit/data/relationships.py
```diff
@@ -163,19 +184,86 @@ def pandas(self, *, attributes: str | list[str] | None = None, ids=False) -> pd.
...
+            )
+
+        if "count" in new_table.column_names:
+            aggregates.append(("count", "sum"))
+            column_renames.update({"count_sum": "count"})
+
+        if aggregates:
+            aggregated_table = table_group.aggregate(aggregates)
+            aggregated_table = aggregated_table.rename_columns(column_renames)
+
```
mdekstrand: A point we forgot: for each of the other columns (not `timestamp`, `count`, or one of the entity number columns), we need an aggregate that takes the `last` value (since we already sort by timestamp). If we have any such aggregates, then we also need to pass `use_threads=False` to `table_group.aggregate`, because paralllel aggregation doesn't work with the `last` aggregate.

src/lenskit/data/relationships.py
```diff
@@ -163,19 +184,86 @@ def pandas(self, *, attributes: str | list[str] | None = None, ids=False) -> pd.
...
+        aggregates: list[tuple[str, str]]
+        column_renames: dict[str, str]
+        aggregates = []
+        column_renames = {}
+
+        if "timestamp" in new_table.column_names:
+            aggregates.extend([("timestamp", "max"), ("timestamp", "min")])
+            column_renames.update(
+                {"timestamp_max": "timestamp", "timestamp_min": "first_timestamp"}
```
mdekstrand, line 260: this raises an interesting question — should the default timestamp be the max or the min?

src/lenskit/data/relationships.py
```diff
@@ -163,19 +184,86 @@ def pandas(self, *, attributes: str | list[str] | None = None, ids=False) -> pd.
...
+            )
+
+        if "count" in new_table.column_names:
+            aggregates.append(("count", "sum"))
+            column_renames.update({"count_sum": "count"})
+
+        if aggregates:
+            aggregated_table = table_group.aggregate(aggregates)
+            aggregated_table = aggregated_table.rename_columns(column_renames)
```
mdekstrand: If this condition is false, we never assign `aggregated_table`, and thus constructing the matrix relationship set will fail. However, we will always at least take a count, so we can get rid of the `if`.

src/lenskit/data/schema.py
```diff
@@ -224,6 +223,10 @@ class RelationshipSchema(BaseModel):
     """
     Relationship attribute definitions.
     """
+    remove_duplicates: bool | Literal["exact"] = False
+    """
+    Whether this relationship should remove repeated interactions or exact duplicates.
+    """
```
mdekstrand: As discussed & noted above, let's not make this a property of the schema.

mdekstrand (changes requested):
Good work! I have quite a few comments, but the overall PR is fundamentally sound, we mostly need to clean up a few things and maintain compatibility.

Where the comments note _behavior_ changes, it would be good to add more tests that exercise the required behavior. It's best to add those tests _before_ updating the code, so you can see them fail because the behavior is wrong, and then pass when you've fixed the behavior.

src/lenskit/data/relationships.py
mdekstrand: This seems to be the correct logic, **but** we will need to make sure that data set splitting does the right thing with time.

src/lenskit/data/builder.py
```diff
@@ -63,6 +63,10 @@
 """
 Action to take when a relationship references a missing entity.
 """
+RepeatOption: TypeAlias = Literal["allow", "forbid", "remove", "remove-duplicate"]
+"""
+Defines how should repeated interactions be handled.
+"""
```
mdekstrand: we don't use this anymore, so let's remove.

src/lenskit/data/builder.py
```diff
@@ -378,6 +381,9 @@ def add_relationships(
                 Whether this is an interaction relationship or not; can be
                 ``"default"`` to indicate this is the default interaction
                 relationship.
+            remove_repeats:
+                If ``True``, repeated interactions will be removed. If ``"exact"``,
+                duplicated interactions will be removed.
```
mdekstrand, line 380: good.

src/lenskit/data/builder.py
```diff
@@ -444,55 +458,59 @@ def add_relationships(
...
-            # add the new data
-            coords.extend(new_table.select(link_nums.keys()).to_batches())
-            n_dupes = len(coords) - coords.unique_count()
-            _log.debug(
-                "coordinates: %d total, %d unique, %d dupe",
-                len(coords),
-                coords.unique_count(),
-                n_dupes,
-            )
+        if remove_repeats and remove_repeats != "exact":
```
mdekstrand: my usual idiom for this is `if remove_repeats is True`.

src/lenskit/data/builder.py
```diff
@@ -444,55 +458,59 @@ def add_relationships(
...
+            new_table = self._remove_repeated_relationships(t=new_table, rel=rc_def)
```
mdekstrand, line 463: The new and current tables haven't been concatenated at this point (that happens on line 517), so this will only resolve repeats within the new data, not repeats between new and old data.

The fact that it passes tests suggests that our tests are not thorough enough.

src/lenskit/data/builder.py
```diff
@@ -444,55 +458,59 @@ def add_relationships(
...
+        else:
+            if remove_repeats == "exact":
+                new_table = self._remove_duplicated_relationships(t=new_table)
```
mdekstrand, line 466: Same as comment above.

src/lenskit/data/builder.py
```diff
@@ -542,10 +561,13 @@ def add_interactions(
                 What to do when interactions reference nonexisting entities; can
                 be ``"error"`` or ``"insert"``.
             allow_repeats:
-                Whether repeated interactions are allowed.
+                Whether repeated interactions are allowed, forbidden, or removed.
```
mdekstrand: We aren't changing this option's semantics anymore.

src/lenskit/data/builder.py
```diff
@@ -1062,6 +1088,35 @@ def save(self, path: str | PathLike[str]):
...
+        _log.debug("checking for repeated interactions")
+        temp_t = t.group_by([entity + "_num" for entity in rel.entity_class_names]).aggregate(
+            [([], "count_all")]
+        )
+        repeat = pc.any(pc.greater(temp_t["count_all"], pa.scalar(1))).as_py()
+        if repeat:
+            if rel.repeats.is_allowed:
+                rel.repeats = AllowableTroolean.PRESENT
+            if rel.repeats.is_forbidden:
+                raise DataError("repeated interactions not allowed for relationship class")
```
mdekstrand: Are we still using this function? If not, remove.

src/lenskit/data/builder.py
```diff
@@ -1074,6 +1129,16 @@ def _resolve_entity_ids(
...
+    def _resolve_repeated_interactions(
+        self, t: pa.Table, rel: RelationshipSchema, remove_repeats: bool | Literal["exact"]
+    ) -> pa.Table:
+        if remove_repeats:
+            if remove_repeats == "exact":
+                t = self._remove_duplicated_relationships(t)
+            else:
+                t = self._remove_repeated_relationships(t, rel)
+        return t
+
```
mdekstrand: Are we still using this method? If not, remove.

src/lenskit/data/relationships.py
```diff
@@ -162,19 +181,96 @@ def pandas(self, *, attributes: str | list[str] | None = None, ids=False) -> pd.
...
+        aggregates = [
+            (col_name, "last")
+            for col_name in new_table.column_names
+            if col_name not in exclude_last
+        ]
+        column_renames = {
+            f"{col_name}_last": col_name
+            for col_name in new_table.column_names
+            if col_name not in exclude_last
+        }
```
mdekstrand, line 251: this is good!

src/lenskit/data/builder.py
mdekstrand: so we need a test that does this:

1. creates a dataset builder
2. adds some interactions
3. adds some more interactions, some of which repeat interactions in (2)
4. builds the data set
5. checks that it is correct

We need to do the same for exact duplicates as well.

Issue #806, opened 2025-07-16
Add efficient lookup and deduplication support for DatasetBuilder


mdekstrand:
The current code to check for duplicate relationships and enforce uniqueness, along with the incoming code for #769, is quite inefficient. This inefficiency compounds when adding relationships in batches.

We can improve this situation by implementing an accelerated index structure in Rust that allows for us to efficiently look up data by entity IDs or by entire rows, and get the locations of that data for replacing existing interactions. It looks like the `arrow::rows` support will be very helpful for this, abstracting away the problem of making rows of tables hashable and comparable. We'll want some wrapper logic, as we don't want to convert the _entire_ relationship table to row-oriented structure, but it looks like we can definitely make it work, with less difficulty than painfully extracting individual values ourselves (at a small efficiency cost).

PR #813, opened 2025-08-15
Refactor indirect hashtables to support reuse

mdekstrand:
This refactors indirect hashtables into their own separate abstraction to facilitate reuse in #806.


mdekstrand:
Turns out we don't actually need this for #806, a different abstraction will work better, but it isn't a bad idea anyway.

PR #850, opened 2025-09-17
Fix: correct attribute-to-entity mapping

samiravaez:
This PR fixes a bug in DatasetBuilder.add_scalar_attribute where scalar attributes were misaligned when added to entity tables.

### DIFF
--- a/src/lenskit/data/builder.py
+++ b/src/lenskit/data/builder.py
@@ -783,11 +783,11 @@ class DatasetBuilder:
             raise DataError(f"{n_bad} unknown entity IDs")
 
         val_array: pa.Array = pa.array(values)  # type: ignore
-        tbl_mask = np.zeros(e_tbl.num_rows, dtype=np.bool_)
-        tbl_mask[nums.to_numpy()] = True
-        tbl_mask = pa.array(tbl_mask)
-        val_col = pa.nulls(e_tbl.num_rows, val_array.type)
-        val_col = pc.replace_with_mask(val_col, tbl_mask, val_array)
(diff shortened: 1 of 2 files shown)


src/lenskit/data/builder.py
```diff
@@ -783,11 +783,11 @@ def add_scalar_attribute(
...
-        tbl_mask = np.zeros(e_tbl.num_rows, dtype=np.bool_)
-        tbl_mask[nums.to_numpy()] = True
-        tbl_mask = pa.array(tbl_mask)
-        val_col = pa.nulls(e_tbl.num_rows, val_array.type)
-        val_col = pc.replace_with_mask(val_col, tbl_mask, val_array)
+        nums_np = nums.to_numpy()
+        vals_np = np.full(e_tbl.num_rows, None, dtype=object)
+        vals_np[nums_np] = val_array.to_numpy(zero_copy_only=False)
+        val_col = pa.array(vals_np, type=val_array.type)
+
```
mdekstrand, line 790: Do we know why the old code is wrong here?

samiravaez: Yes, the old code assumed that the input attributes were ordered by entity number. When the input DataFrame had IDs out of order, the replacement wrote values into the wrong positions. Then, when converting back to pandas, the index (entity IDs) appeared correct, but the attribute values were mismatched. The new code scatters the values into a numpy array aligned to the entity numbers

mdekstrand: Thanks! Now that you point it out, I see it too. Good catch!

Issue #778, opened 2025-05-29
Add reranking behavior metric functions


mdekstrand:
To support measuring the behavior of reranking components, it would be useful to add functions that will compute Rank Biased Overlap and Least Item Promoted given two lists (a base ranking and reranked outputs).

For the initial task (this issue), these should just be functions that take two `ItemList`s along with other parameters (i.e., the weighting model), and return the score.

When working on #766, we can work on an interface to incorporate these kinds of metrics into an analysis. Getting the code into LensKit as functions without integration into an analysis or evaluation loop will allow the code to at least be reused in prototype experiments while we work on the harder interface design problem.

PR #856, opened 2025-10-02
Add reranking behavior metrics: rbo and lip

sushobhan2024:
This covers issue #778

### DIFF
--- a/src/lenskit/metrics/reranking/lip.py
+++ b/src/lenskit/metrics/reranking/lip.py
@@ -1,0 +1,48 @@
+# This file is part of LensKit.
+# Copyright (C) 2018-2023 Boise State University.
+# Copyright (C) 2023-2025 Drexel University.
+# Licensed under the MIT license, see LICENSE.md for details.
+# SPDX-License-Identifier: MIT
+
+from __future__ import annotations
+
+import numpy as np
+
+from lenskit.data import ItemList
+
+
+def least_item_promoted(reference: ItemList, reranked: ItemList, k: int = 10) -> float:
+    """
+    Compute the Least Item Promoted (LIP) metric.
+
+    This metric identifies the item in the top-k reranked list that had the
+    highest (worst) rank in the reference, and returns how many positions
+    it was promoted from beyond k.
+
+    Args:
+        reference:
+            The original/base ranking.
+        reranked:
+            The reranked list.
+        k:
+            The depth to evaluate (default 10).
+
+    Returns:
+        The rank distance of the least-promoted item, or NaN if the base
+        ranking is empty.
+    """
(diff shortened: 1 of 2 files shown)


src/lenskit/metrics/reranking/_rbo.py
```diff
@@ -0,0 +1,48 @@
...
+        The RBO score between 0 and 1.
+    """
+    rank_weight = GeometricRankWeight(p)
+    weights = rank_weight.weight(np.arange(1, k + 1))
+    sum = 0
+    total_weights = 0
+    for d, weight in enumerate(weights, start=1):
+        set_a = set(reference[:d].ids())
+        set_b = set(reranked[:d].ids())
+        overlap = len(set_a.intersection(set_b))
```
mdekstrand: can you do this with `np.intersect1d` instead of going into `set`?

src/lenskit/metrics/reranking/_rbo.py
```diff
@@ -0,0 +1,48 @@
...
+    ACM Transactions on Information Systems (TOIS) 28.4 (2010): 20.
+    https://dl.acm.org/doi/10.1145/1852102.1852106
+
+    Args:
+        reference:
+            The first item list to compare.
+        reranked:
+            The second item list to compare.
+        p:
+            The patience parameter (default 0.85).
```
mdekstrand: Should we make the function take a whole weight object instead of just the patience parameter?

tests/eval/test_rbo.py
```diff
@@ -0,0 +1,53 @@
...
+def test_rbo_empty_reference():
+    ref = ItemList()
+    rerank = ItemList([1, 2, 3], ordered=True)
+    assert call_metric(RBO, ref, rerank) == 0.0
+
+
+def test_rbo_empty_rerank():
+    ref = ItemList([1, 2, 3], ordered=True)
+    rerank = ItemList()
+    assert call_metric(RBO, ref, rerank) == 0.0
```
mdekstrand: Very good!

src/lenskit/metrics/reranking/_rbo.py
```diff
@@ -0,0 +1,48 @@
...
+
+from __future__ import annotations
+
+import numpy as np
+
+from lenskit.data import ItemList
+from lenskit.metrics import GeometricRankWeight
+
+
+def RBO(reference: ItemList, reranked: ItemList, p: float = 0.85, k: int = 10) -> float:
```
mdekstrand: Since these are functions, let's name them lower-case, possibly snake-case, to save the uppercase names for future metric classes.

Possibly `rank_biased_overlap` and `least_item_promoted`? No harm in a full-length name.

PR #860, opened 2025-10-03
Add Fair Re-Ranker (FA*IR)

samiravaez:
This PR adds the FA*IR reranker, based on Zehlike et al. (2017), and integrates it into the `RecPipelineBuilder`.

### DIFF
--- a/src/lenskit/pipeline/common.py
+++ b/src/lenskit/pipeline/common.py
@@ -38,6 +38,7 @@ class RecPipelineBuilder:
     is_predictor: bool = False
     _predict_transform: Component | None = None
     _fallback: Component | None = None
+    _reranker: CompRec | None = None
 
     def __init__(self):
         from lenskit.basic.candidates import UnratedTrainingItemsCandidateSelector
@@ -144,10 +145,36 @@ class RecPipelineBuilder:
         rank = pipe.add_component(
             "ranker", self._ranker.component, self._ranker.config, items=n_score, n=n_n
         )
-        pipe.alias("recommender", rank)
-        pipe.default_component("recommender")
+
+        # If a reranker is configured, attach it
+        if self._reranker is not None:
+            rerank = pipe.add_component(
+                "reranker",
+                self._reranker.component,
+                self._reranker.config,
+                items=rank,
+            )
+            pipe.alias("recommender", rerank)
+            pipe.default_component("recommender")
+        else:
+            pipe.alias("recommender", rank)
+            pipe.default_component("recommender")
 
         return pipe.build()
+
+    def reranker(self, kind: str, **params):
+        """
+        Specify a reranker to apply after ranking.
+        Example:
+            builder.reranker("fair", k=10, p=0.5, alpha=0.1)
+        """
+        from lenskit.reranking.fair import FairReranker
+
+        if kind == "fair":
+            cfg = FairReranker.validate_config(params)
+            self._reranker = CompRec(FairReranker, cfg)
+        else:
+            raise ValueError(f"unknown reranker type {kind}")
 
 
 def topn_builder(
--- a/src/lenskit/reranking/fair.py
+++ b/src/lenskit/reranking/fair.py
@@ -1,0 +1,223 @@
+# This file is part of LensKit.
+# Copyright (C) 2018-2023 Boise State University.
+# Copyright (C) 2023-2025 Drexel University.
+# Licensed under the MIT license, see LICENSE.md for details.
+# SPDX-License-Identifier: MIT
+
+"""
+FA*IR top-k re-ranking.
+"""
+
+# pyright: basic
+from __future__ import annotations
+
+from collections import deque
+
+import numpy as np
+from pydantic import BaseModel, Field, PositiveFloat, PositiveInt
+from scipy.stats import binom
+from typing_extensions import override
+
+from lenskit.data import Dataset, ItemList
+from lenskit.logging import Stopwatch, get_logger
+from lenskit.pipeline import Component
+from lenskit.training import Trainable, TrainingOptions
+
+_log = get_logger(__name__)
+
+
+class FairRerankerConfig(BaseModel, extra="forbid"):
+    """
+    Configuration for :class:`FairTopNReranker`.
+    """
+
+    k: PositiveInt | None = None
+    """
+    Size of the top-k segment to rerank. If ``None``, the entire
+    candidate list is reranked (unbounded top-k).
+    """
+
+    p: PositiveFloat = Field(0.5, gt=0.0, lt=1.0)
+    """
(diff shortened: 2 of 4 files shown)


src/lenskit/reranking/fair.py
```diff
@@ -0,0 +1,223 @@
...
+
+_log = get_logger(__name__)
+
+
+class FairRerankerConfig(BaseModel, extra="forbid"):
+    """
+    Configuration for :class:`FairTopNReranker`.
+    """
+
+    k: PositiveInt | None = None
```
mdekstrand, line 34: The standard name for this in LensKit is `n`, and we generally use the term “top-*N*” (some people use top-*K*, others top-*N* as their standard termology; LensKit uses top-*N*).

src/lenskit/reranking/fair.py
```diff
@@ -0,0 +1,223 @@
...
+        k:
+            Size of the top-k subset to rerank. If ``None``, the entire
+            candidate list is reranked.
+        p:
+            The target proportion of protected items in the list (0.0 ≤ p ≤ 1.0).
+            The default value is 0.5.
+        alpha:
+            The statistical significance level for the fairness test.
+            The default value is 0.1.
+
```
mdekstrand, line 77: We don't need to duplicate this documentation here, the configuration class is sufficient.

src/lenskit/reranking/fair.py
```diff
@@ -0,0 +1,223 @@
...
+        Args:
+            protected attributes:
+                The item entity table must have a scalar attribute ``protected`` that is ``True``
+                if the item is in the protected group and ``False`` otherwise.
+        """
+        self.pmf_cache = {}
+
+        log = _log.bind(alpha=self.config.alpha, k_max=self.config.k)
+        timer = Stopwatch()
+        log.info("begining FA*IR Tuning")
```
mdekstrand, line 163: we aren't really tuning anything, are we? Maybe “computing FA*IR thresholds” or something like that?

src/lenskit/reranking/fair.py
```diff
@@ -0,0 +1,223 @@
...
+        p_items = deque()
+        up_items = deque()
+
+        is_protected = self.protected_attributes.reindex(item_ids).fillna(False).values
+
+        p_items = deque(np.nonzero(is_protected)[0])
+        up_items = deque(np.nonzero(~is_protected)[0])
+
+        count_prot = 0
+        reranked_indices = deque()
```
mdekstrand, line 202: since we're only appending, not popping from the front, we can just use a list here instead of a `deque`.

src/lenskit/reranking/fair.py
```diff
@@ -0,0 +1,223 @@
...
+                    self.pmf_cache[block_size] = binom.pmf(np.arange(block_size + 1), block_size, p)
+
+                S = np.convolve(self.pmf_cache[block_size], S)
+                # set the next index (j-1) to zero since the last element of each block
+                # increases by 1 and now we need at least j protected up to this point
+                S[j - 1] = 0
+
+        return float(1 - S.sum())
+
+    def binary_search_significance(self, k, p, alpha, tolerance=1e-10, max_iter=100):
```
mdekstrand, line 128: Since this is internal, let's prefix it with `_`.

src/lenskit/reranking/fair.py
```diff
@@ -0,0 +1,223 @@
...
+        )
+        # 2. compute m_table
+        self.m_list_ = self._compute_m_list(k=self.config.k, p=self.config.p, alpha=self.alpha_c_)
+
+        items = data.entities("item")
+
+        if "protected" not in items.attributes:
+            raise ValueError("Dataset is missing required 'protected' attribute for item entities")
+
+        self.protected_attributes = items.attribute("protected").pandas()
```
mdekstrand, line 177: Do we want to make the attribute name configurable?

src/lenskit/pipeline/common.py
```diff
@@ -144,11 +145,37 @@ def build(self, name: str | None = None) -> Pipeline:
...
+            )
+            pipe.alias("recommender", rerank)
+            pipe.default_component("recommender")
+        else:
+            pipe.alias("recommender", rank)
+            pipe.default_component("recommender")
 
         return pipe.build()
 
+    def reranker(self, kind: str, **params):
```
mdekstrand, line 165: This should work like `ranker`, `candidate_selector`, etc.: it should take the component class or object, and an optional `config` object, rather than names — the code in the `pipeline` package generally shouldn't know anything about specific components in the rest of LensKit.

src/lenskit/reranking/fair.py
```diff
@@ -0,0 +1,223 @@
...
+    The target proportion of protected items in the list (0.0 ≤ p ≤ 1.0).
+    """
+
+    alpha: PositiveFloat = Field(0.1, gt=0.0, lt=1.0)
+    """
+    Family-wise Type I error across prefixes.
+    """
+
+
+class FairReranker(Component[ItemList], Trainable):
```
mdekstrand, line 51: Since the algorithm is called "FA*IR", maybe call this `FAIRReranker`?

mdekstrand (changes requested):
Good work! A few changes here for consistency, but I don't see any obvious problems with the logic.

One thing that we should perhaps discuss and decide: this version treats the items with unknown protected attributes as unprotected. An alternate variant would be to leave them as unknown, and mix them in to the final ranking.

samiravaez:
> Good work! A few changes here for consistency, but I don't see any obvious problems with the logic.
> 
> One thing that we should perhaps discuss and decide: this version treats the items with unknown protected attributes as unprotected. An alternate variant would be to leave them as unknown, and mix them in to the final ranking.

I don’t think it makes much difference. We only track and tune the female distribution. For the rest (unknown or unprotected) the algorithm always takes the top-scoring item for the next position, so I think it leads to the same outcome.

src/lenskit/reranking/fair.py
samiravaez: Sure, I just thought `k` might be used in the reranking context, since we’re dealing with a subset of `n`.

src/lenskit/reranking/fair.py
samiravaez: Sure. Should I set the default value to `protected`?

src/lenskit/reranking/fair.py
mdekstrand: I think it's clearer to just talk about “top-N”, and allow for different stages with different Ns. Then we don't have to remember when to use N and when to use K, or what order they might go in, etc.

src/lenskit/reranking/fair.py
mdekstrand: That sounds good.

PR #890, opened 2025-10-16
Add entropy and rank biased entropy

sushobhan2024:
Added a function `matrix_column_entropy` that takes a matrix and optional weight vector and computes entropy. `entropy` and `rank_biased_entropy` functions use this `matrix_column_entropy`. 

This PR does not cover creating ENTROPY and RBENTROPY metric classes. Once this gets merged, I will write metric classes that handle the conversion of category into matrix.

### DIFF
--- a/src/lenskit/metrics/ranking/_entropy.py
+++ b/src/lenskit/metrics/ranking/_entropy.py
@@ -90,10 +90,7 @@ def matrix_column_entropy(
         return np.nan
 
     if weights is not None:
-        if sps.issparse(matrix):
-            matrix = matrix.multiply(weights[:, np.newaxis])
-        else:
-            matrix = np.multiply(matrix, weights[:, np.newaxis])
+        matrix = matrix * weights[:, np.newaxis]
 
     values = np.asarray(matrix.sum(axis=0))
     values = values + 1e-6
--- a/tests/eval/test_entropy.py
+++ b/tests/eval/test_entropy.py
@@ -22,7 +22,7 @@ def test_entropy_uniform():
     row = [0, 1, 2]
     col = [0, 1, 2]
     data = [1, 1, 1]
-    sparse = sps.csr_matrix((data, (row, col)), shape=(3, 3))
+    sparse = sps.csr_array((data, (row, col)), shape=(3, 3))
 
     dense_result = entropy(items, dense)  # n is lenght of items i.e. 3
     sparse_result = entropy(items, sparse, n=5)  # n is max(5, len(items)) i.e. 3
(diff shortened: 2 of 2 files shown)


sushobhan2024:
- Handled sparse and dense matrices separately w/o conversion sparse to dense
- Included zeros and added smoothing 
- Added tests for `matrix_column_entropy()`
- Created sparse matrices in test using spicy `csr_matrix((data, (row_ind, col_ind)), [shape=(M, N)])`

src/lenskit/metrics/ranking/_entropy.py
```diff
@@ -0,0 +1,108 @@
...
+    from lenskit.metrics import GeometricRankWeight
+
+    if n is None:
+        n = len(items)
+    if n == 0:
+        return np.nan
+
+    n = min(n, len(items))
+
+    truncated = categories[:n, :] if sps.issparse(categories) else categories[:n]
```
mdekstrand: Should be able to use `categories[:n, :]` in both cases, no?

src/lenskit/metrics/ranking/_entropy.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+
+    if matrix.shape[0] == 0 or matrix.shape[1] == 0:
+        return np.nan
+
+    if weights is not None:
+        if sps.issparse(matrix):
+            matrix = matrix.multiply(weights[:, np.newaxis])
+        else:
+            matrix = matrix * weights[:, np.newaxis]
```
mdekstrand: can you use `.multiply` in both cases?

src/lenskit/metrics/ranking/_entropy.py
```diff
@@ -0,0 +1,108 @@
...
+
+    if sps.issparse(matrix):
+        values = np.asarray(matrix.sum(axis=0)).ravel()
+    else:
+        values = matrix.sum(axis=0)
```
mdekstrand: this should also be able to be done with the same logic (`np.asarray` is fine to call on something that's already an array).

mdekstrand (commented):
Some simplifications seem possible, then we can merge.

src/lenskit/metrics/ranking/_entropy.py
sushobhan2024: yes, it works for both cases.

src/lenskit/metrics/ranking/_entropy.py
sushobhan2024: I used `np.multiply()` for the dense array.

src/lenskit/metrics/ranking/_entropy.py
sushobhan2024: Am I using `.ravel()` correctly here? For the dense matrix `.ravel()` is not necessary, but for sparse matrix.

mdekstrand: I just tested in an interactive session, and `.sum(axis=0)` on a `csr_array` returns a dense array with the correct shape, so no special handling should be needed.

PR #907, opened 2025-10-26
Add support for dataframe and arrow table in add_entities

FroggoLight:
Allows client to call add_entities with either dataframe or arrow table without having to call individual add_attributes for all of the data columns.

### DIFF
--- a/src/lenskit/data/builder.py
+++ b/src/lenskit/data/builder.py
@@ -254,16 +254,27 @@ class DatasetBuilder:
         """
         Add entities to the data set.
 
+        When constructed with a data frame or table, this method looks for
+        entitiy IDs in the column ``{cls}_id`` (e.g. ``item_id``).  If no such
+        column exists, and the data frame is a Pandas data frame, then item IDs
+        are taken from the data frame's index.  A warning is issued if the data
+        frame index name is not ``{cls}_id``.
+
         Args:
             cls:
                 The name of the entity class (e.g. ``"item"``).
             source:
-                The input data, as an array or list of entity IDs.
-
-                .. note::
-                    Data frame support will be added in a future version.
+                The input data, as an array or list of entity IDs, or a data
+                frame of entities with attributes.
             duplicates:
                 How to handle duplicate entity IDs.
+        Raises:
+            DataError:
+                When there is a fatal problem with the supplied data.
+        Warns:
+            DataWarning:
+                When the data is valid but suspect, such as a data frame with no
+                column or index named ``{cls}_id``.
         """
         id_col = f"{cls}_id"
         if isinstance(source, pd.DataFrame):


tests/data/test_builder_entities.py
```diff
@@ -143,3 +145,38 @@ def test_add_entities_twice():
...
+
+    genres = items["genres"].str.split("|")
+    items["genres"] = genres
+
+    dsb.add_entities("item", items)
+
+    ds = dsb.build()
+
+    assert ds.entities("item").attribute("title").is_scalar
+    assert ds.entities("item").attribute("genres").is_list
```
mdekstrand, line 179: We should test that a few item IDs have the correct titles, too. It's possible for the code to set up the structures in the right format, but not align them correctly, and the tests should check for that.

tests/data/test_builder_entities.py
```diff
@@ -143,3 +145,38 @@ def test_add_entities_twice():
...
+    genres = items["genres"].str.split("|")
+    items["genres"] = genres
+    table = pa.Table.from_pandas(items)
+
+    dsb.add_entities("item", table)
+
+    ds = dsb.build()
+
+    assert ds.entities("item").attribute("title").is_scalar
+    assert ds.entities("item").attribute("genres").is_list
```
mdekstrand, line 203: Same as above.

src/lenskit/data/builder.py
```diff
@@ -265,10 +265,23 @@ def add_entities(
             duplicates:
                 How to handle duplicate entity IDs.
         """
-        if isinstance(source, pd.DataFrame):  # pragma: nocover
-            raise NotImplementedError()
-        if isinstance(source, pa.Table):  # pragma: nocover
-            raise NotImplementedError()
+        if isinstance(source, pd.DataFrame):
+            source = pa.Table.from_pandas(source)
```
mdekstrand: There is an interesting and challenging edge case here, that we need to clearly document and/or design for.

Right now, this works because your test case names the index `item_id`, which then turns into a column when we do `from_pandas`.

However, if the client provides code that has no `item_id` column, and has an index with a different name, we need to figure out what to do. Do we want to use the index? Do we want to throw an error?

I think we _probably_ want to use the Pandas index, with the following logic:

1. If the data frame has a column named `{cls}_id`, use that column as the entity IDs, and ignore the index.
2. Otherwise, assume the index has entity IDs.

Implementing this logic will require this line to be a little more aware of the Pandas data frames, and also require tests for each of the different conditions. Importantly, for case (1), this line here will create a new attribute called `index` (or whatever the index name is), and we don't want to include that.

The input cases we will need to test for correct behavior with:

1. Input has an index named `{cls}_id` (the current test)
2. Input has an index named something else, and no column named `{cls}_id`
3. Input has a column named `{cls}_id`

This isn't a problem for PyArrow input, because Arrow tables do not have indices.

src/lenskit/data/builder.py
```diff
@@ -265,10 +265,23 @@ def add_entities(
...
+        if isinstance(source, pa.Table):
+            entity_col = source.column(cls + "_id")
+            self.add_entities(cls, entity_col)
+
+            for col_name in source.column_names:
+                if not col_name.endswith("_id"):
```
mdekstrand: We should only exclude the `{cls}_id` column — we want any other `_id` columns to result in an error, not be silently ignored.

src/lenskit/data/builder.py
```diff
@@ -265,10 +265,23 @@ def add_entities(
             duplicates:
```
mdekstrand, line 269: We should update the docstring to document the kinds of attributes supported, limitations, etc., along with the index logic.

This should be in the main body of the docstring (before `Args:`), not in the argument documentation, for readability.

Issue #892, opened 2025-10-16
Add category matrix support to AttributeSet


mdekstrand:
As discussed today with @sushobhan2024, we need to add a new `cat_matrix` method to `AttributeSet` and its subclasses, to compute a matrix over categorical values for various attribute representations.

- [ ] Define the `cat_matrix` signature. Proposed:

    ```python
    def cat_matrix(self, *, normalize: Literal["unit" | "distribution"] = None) -> NDArray[np.floating[Any]] | csr_array: ...
    ```

- [ ] Implement `cat_matrix` for scalar attribute sets, where the columns of the matrix refer to distinct values of the attribute. It should throw a `TypeError` if the attribute is floating-point. We may want to cache a `Vocabulary` of distinct values to speed up repeated calls. If the attribute is null for a given item, the resulting matrix should be all 0. Should usually return a sparse matrix (`csr_array`).

- [ ] Implement `cat_matrix` for list attribute sets. Behaves like scalar attribute, except that more than one value can be nonzero.

- [ ] Implement `cat_matrix` for dense vector attribute sets. Returns a dense matrix (`np.ndarray`), columns correspond to the attribute vector columns. Basically just returns the vector as-is, possibly with normalized rows.

- [ ] Implement `cat_matrix` for sparse vector attribute sets. Same as dense, except it returns a sparse matrix instead of a dense one.

Open question: do we provide access to the vocabulary for list and scalar attribute sets, and how? Maybe we also add a `cat_vocabulary()` method that returns the column vocabulary for a vector attribute set, and the unique value vocabulary for a scalar or list set, and `None` if the scalar or list is floating-point or if the vector attribute has no dimension vocabulary / names.

PR #912, opened 2025-10-30
Add `cat_matrix` to `AttributeSet`

sushobhan2024:
#892

- [x] Define the cat_matrix signature.
- [x] Implement cat_matrix for scalar and list attribute sets
- [x] Implement cat_matrix for dense vector and sparse vector attribute sets
- [x] Write tests for cat_matrix and normalize_matrix


src/lenskit/data/matrix.py
```diff
@@ -542,3 +543,46 @@ def _check_index_type(index_type: pa.DataType, dimension: int | None) -> int:
...
+        Normalized matrix
+    """
+    if normalize is None:
+        return matrix
+
+    is_sparse = isinstance(matrix, csr_array)
+
+    if normalize == "unit":
+        if is_sparse:
+            stats = np.sqrt(np.asarray(matrix.multiply(matrix).sum(axis=1)).ravel())
```
mdekstrand: this can be done with `scipy.sparse.linalg.norm`

src/lenskit/data/matrix.py
```diff
@@ -542,3 +543,46 @@ def _check_index_type(index_type: pa.DataType, dimension: int | None) -> int:
...
+            stats = np.asarray(matrix.sum(axis=1)).ravel()
+        else:
+            stats = matrix.sum(axis=1)
+
+    stats[stats == 0] = 1.0
+
+    if is_sparse:
+        matrix = matrix.copy()
+        matrix.data = matrix.data.astype(np.float32)
+        matrix.data /= np.repeat(stats, np.diff(matrix.indptr))
```
mdekstrand: can this be done with `matrix / stats[:, None]` like the dense version? If not, does the `.divide` method work?

src/lenskit/data/attributes.py
```diff
@@ -130,6 +130,14 @@ def numbers(self) -> np.ndarray[tuple[int], np.dtype[np.int32]]:
         else:
             return self._selected.to_numpy()
 
+    def cat_matrix(
+        self, *, normalize: Literal["unit", "distribution"] | None = None
+    ) -> tuple[NDArray[np.floating] | csr_array, Vocabulary | None]:
+        """
+        Compute a categorical matrix representation of the attribute.
```
mdekstrand, line 137: Let's explain what this means.

tests/data/test_attribute_set.py
```diff
@@ -0,0 +1,127 @@
...
+    ScalarAttributeSet,
+    SparseAttributeSet,
+    VectorAttributeSet,
+)
+from lenskit.data.matrix import SparseRowArray
+from lenskit.data.schema import AttrLayout, ColumnSpec
+from lenskit.data.vocab import Vocabulary
+
+
+def test_scalar():
```
mdekstrand: since these are testing categorical matrices, let's call these things like `test_scalar_cat_matrix`.

mdekstrand: in addition to test names, might be good to call this whole file `test_attribute_cat_matrix.py`, and other attribute set tests can go in other files, to keep any one file from getting too big

src/lenskit/data/matrix.py
sushobhan2024: `matrix / stats[:, None]` for sparse matrix returns `coo_array` then, in the updated code, I am converting it to `csr_array` with `.tocsr()`.

PR #918, opened 2025-10-31
FA*IR: bug fixes and test updates

samiravaez:
This PR fixes a few bugs and adds additional tests

### DIFF
--- a/conda/recipe.yaml
+++ b/conda/recipe.yaml
@@ -36,7 +36,7 @@ outputs:
       run:
         - python >=${{ python_min }}
         - typing-extensions >=4.12,<5
-        - pyarrow >=20,<21
+        - pyarrow >=20
         - pandas >=2.0,<3
         - numpy >=1.25,<3
         - scipy >=1.11,<2


mdekstrand:
What if we used NumPy instead of Pandas, and saved the data set's `items` vocabulary along with a NumPy `protected` array?

In `__call__`, you can then get the numbers for the item list with `items.numbers(vocabulary=self.items, missing="negative")`, and use those to look up the protected attributes.

That would remove the need for the Pandas cast warning option, which seems to be causing our remaining test failures.

samiravaez:
> What if we used NumPy instead of Pandas, and saved the data set's `items` vocabulary along with a NumPy `protected` array?
> 
> In `__call__`, you can then get the numbers for the item list with `items.numbers(vocabulary=self.items, missing="negative")`, and use those to look up the protected attributes.
> 
> That would remove the need for the Pandas cast warning option, which seems to be causing our remaining test failures.

I made the changes, but there’s still a failing test; I think it’s not from the reranker.

Issue #908, opened 2025-10-27
Add flexible weight inputs for RBP


mdekstrand:
To support external innovation in evaluation metrics and weights, we need to update RBP (and perhaps related metrics) to allow client code to directly provide weights. There are a few pieces to enable this, similar to the design we're using for entropy metrics in #890:

- [ ] Add a pure function, `rank_biased_precision`, to compute RBP given explicit weights and a normalization factor. This will allow a lot more experimentation in client code to reuse LensKit's (tested) RBP implementation, allowing flexibility without proliferating metric implementations.
- [ ] Modify the `RBP` metric to use this new function.
- [ ] Update the `RBP` metric to allow the user to specify a field name for the weights, in addition to a rank model; this field name is read from the input item lists. This will allow client code to use the metric & measurement collector interfaces with custom weights.

This is just for RBP right now, but we probably want similar changes for other metrics that use rank weights.

mdekstrand:
Fixed in #923.

PR #923, opened 2025-11-04
Add `rank_biased_precision` function that takes explicit weights

sushobhan2024:
This PR covers issue #908.

- [x] Added a pure function, `rank_biased_precision`, to compute RBP given explicit weights and a normalization factor.
- [x] Modified the RBP metric to use this new function.
- [x] Updated the RBP metric to allow the user to specify a field name `weight_field` for the weights, in addition to a rank model.
- [x] Added tests to check `weight_field` and `rank_biased_precision` work.


src/lenskit/metrics/ranking/_rbp.py
```diff
@@ -106,24 +130,36 @@ def measure_list(self, recs: ItemList, test: ItemList) -> float:
...
-            # normalize by max achieveable RBP
-            max = np.sum(weight[: min(nrel, k)]).item()
-            return rbp / max
-        elif wmax is not None:
-            # normalization defined by metric
-            return rbp / wmax
+        if self.weight_field is not None:
+            # use custom weights from field
+            weights = recs.field(self.weight_field)
+            weights = np.asarray(weights)
```
mdekstrand: `weights` is already an array.

src/lenskit/metrics/ranking/_rbp.py
```diff
@@ -106,24 +130,36 @@ def measure_list(self, recs: ItemList, test: ItemList) -> float:
...
-            # normalize by total weight
-            all_weight = self.weight.weight(ranks)
-            return rbp / np.sum(all_weight).item()
+            ranks = recs.ranks()
+            assert ranks is not None
+
+            # we only need to sum good weights
+            good_ranks = ranks[good]
+            weight = self.weight.weight(good_ranks)
+
```
mdekstrand: With the new `rank_biased_precision` function, perhaps messing around with limiting to "good" ranks is an unnecessary optimization. It certainly makes this code more complicated, and I'm not sure that complexity is worth it.

src/lenskit/metrics/ranking/_rbp.py
```diff
@@ -13,6 +13,27 @@
...
+    """
+    Compute rank-biased precision given explicit weights.
+
+    Args:
+        good:
+            Boolean array indicating relevant items at each position.
+        weights:
+            Weight for each item position (same length as good).
+        normalization:
+            Normalization factor.
```
mdekstrand: Normalization should be optional, and default to 1.0.

PR #957, opened 2025-11-19
Resolves #956: Fix NDCG ZeroDivisionError

### LINKED ISSUE #956: NDCG crashes with ZeroDivisionError for users with empty test set
NDCG fails with a ZeroDivisionError when a user's test list is empty.

samiravaez:
This PR resolves #956. It updates NDCG to safely handle cases where the ideal DCG is zero by returning NaN instead of raising ZeroDivisionError

### DIFF
--- a/src/lenskit/metrics/ranking/_dcg.py
+++ b/src/lenskit/metrics/ranking/_dcg.py
@@ -96,6 +96,9 @@ class NDCG(ListMetric, RankingMetricBase):
     def measure_list(self, recs: ItemList, test: ItemList) -> float:
         recs = self.truncate(recs)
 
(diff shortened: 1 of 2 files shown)


src/lenskit/metrics/ranking/_dcg.py
```diff
@@ -96,6 +96,9 @@ def label(self):
     def measure_list(self, recs: ItemList, test: ItemList) -> float:
         recs = self.truncate(recs)
 
+        if len(test) == 0:
+            return float("nan")
```
mdekstrand, line 100: clearer IMO to return `math.nan` or `np.nan` instead of going through a string.

Issue #317, opened 2022-07-27
Add intra-list similarity diversity metric


mdekstrand:
It would be useful to add an intra-list similarity metric for computing diversity to LensKit. As with #316, this metric will need additional data (the item vectors).

Other metrics can also be considered (intra-list distance, entropy).

PR #969, opened 2025-12-08
Add ILS

sushobhan2024:
Issue: https://github.com/lenskit/lkpy/issues/317
Added ILS and tests.

### DIFF
--- a/src/lenskit/metrics/__init__.py
+++ b/src/lenskit/metrics/__init__.py
@@ -68,14 +68,9 @@ __all__ = [
     "quick_measure_model",
     "least_item_promoted",
     "rank_biased_overlap",
-<<<<<<< feature/ILS
-    "entropy",
-    "rank_biased_entropy",
     "ILS",
-=======
     "Entropy",
     "RankBiasedEntropy",
->>>>>>> main
 ]
 
 P = ParamSpec("P")
--- a/src/lenskit/metrics/ranking/__init__.py
+++ b/src/lenskit/metrics/ranking/__init__.py
@@ -40,4 +40,5 @@ __all__ = [
     "ExposureGini",
     "Entropy",
     "RankBiasedEntropy",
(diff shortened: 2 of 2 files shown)


src/lenskit/metrics/ranking/_ils.py
```diff
@@ -0,0 +1,104 @@
...
+
+    return float(ils_score)
+
+
+class ILS(ListMetric, RankingMetricBase):
+    """
+    Evaluate recommendation diversity using intra-list similarity (ILS).
+
+    This metric measures the average pairwise cosine similarity between item
+    category vectors in a recommendation list. Lower values indicate more diverse
```
mdekstrand: ILS should not be limited to category vectors - it should also support arbitrary vectors (which is sort-of already the case, since `cat_matrix` returns the vectors as-is from sparse and dense attribute vectors), but the documentation should reflect the metric's generality.

src/lenskit/metrics/ranking/_ils.py
```diff
@@ -0,0 +1,104 @@
...
+from ._base import ListMetric, RankingMetricBase
+
+
+def intra_list_similarity(items: ItemList, categories: np.ndarray | sps.spmatrix) -> float:
+    """
+    Compute intra-list similarity between item category vectors.
+
+    Args:
+        items: Item list to evaluate.
+        categories: Item * category matrix (dense or sparse), unit-normalized.
```
mdekstrand: Related to the other comment on categories vs. other names, let's call this `vectors`, and otherwise update the docs to reflect that we use vectors.

mdekstrand: Document that the _rows_ of the matrix must be unit-normalized.

mdekstrand (commented):
the code looks sound, some documentation comments (and we need to fix merge conflicts).

PR #983, opened 2026-01-22
Metric interface clean up

sushobhan2024:
As we discussed, I -- 

1. removed `DecomposedMetric` class,
2. kept only `measure_run` method in `GlobalMetric` class, and avoided extending `Metric` class.

### DIFF
--- a/docs/releases/2026.rst
+++ b/docs/releases/2026.rst
@@ -21,6 +21,11 @@ Breaking Changes
 - LensKit now requires Python 3.12 or newer, along with NumPy 2.x, Pandas 2.3 or
   newer, and SciPy 1.13 or newer (see :ref:`dep-policy`, :pr:`954`).
 - We no longer publish 32-bit binary wheels.
+- Removed ``DecomposedMetric``, as the :class:`~lenskit.metrics.Metric` interface
+  is now decomposed. All metrics based on listwise measurements or intermediate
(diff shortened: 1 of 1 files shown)


src/lenskit/metrics/ranking/_gini.py
```diff
@@ -16,13 +16,13 @@
...
 
-from ._base import DecomposedMetric, RankingMetricBase
+from ._base import RankingMetricBase
 from ._weighting import GeometricRankWeight, RankWeight
 
 _log = get_logger(__name__)
 
 
-class GiniBase(DecomposedMetric, RankingMetricBase):
+class GiniBase(RankingMetricBase):
```
mdekstrand, line 25: Does `RankingMetricBase` inherit from `ListMetric`? Gini is not a listwise metric, so if it does then we should either:

- Make `RankingMetricBase` no longer inherit from `ListMetric`, and update all the ranking metrics that _are_ listwise metrics to inherit from both `RankingMetricBase` and `ListMetric`.
- Rename `RankingMetricBase` to `RankingListMetricBase` and make `GiniBase` no longer inherit from `RankingMetricBase`
- A combination of the two: introduce `RankingListMetricBase` to be both ranking metric and list metric, and make the other metrics inherit from it.

sushobhan2024: `RankingMetricBase` inherits from `Metric`.

PR #1012, opened 2026-02-21
Add EASE scorer

### LINKED ISSUE #942: Add EASE
It would be useful to add the [EASE model](https://dl.acm.org/doi/10.1145/3308558.3313710), widely used as a baseline, to LensKit.

mdekstrand:
This adds an implementation of [EASE][], along with supporting changes to matrix relationship sets and co-occurrence counting.

It could be further sped up (a little) by directly counting into a dense matrix, instead of counting into a sparse matrix and then making it dense.

Closes #942.

[EASE]: https://dl.acm.org/doi/10.1145/3308558.3313710

### DIFF
--- a/src/lenskit/knn/ease.py
+++ b/src/lenskit/knn/ease.py
@@ -73,17 +73,7 @@ class EASEScorer(Component[ItemList], Trainable):
     Item interpolation weight matrix.
     """
 
-    @staticmethod
-    def is_available() -> bool:
-        return Version(scipy.__version__) >= MIN_SCIPY_VERSION
-
     def train(self, data: Dataset, options: TrainingOptions | None = None):
-        if not self.is_available():
-            _log.error(
-                "scipy version %s installed, EASE requires %s", scipy.__version__, MIN_SCIPY_VERSION
-            )
-            raise RuntimeError(f"EASE requires SciPy {MIN_SCIPY_VERSION} or later")
-
         if options is None:
             options = TrainingOptions()
 
@@ -142,6 +132,8 @@ class EASEScorer(Component[ItemList], Trainable):
             case "torch":
                 mat = _chol_invert_torch(cooc, device=options.configured_device(gpu_default=True))
             case "scipy":
+                if Version(scipy.__version__) < MIN_SCIPY_VERSION:
+                    raise RuntimeError("SciPy solver requires SciPy 1.17 or later")
                 mat = _chol_invert_scipy(cooc)
         log.info("inverted co-occurrance matrix in %s", timer)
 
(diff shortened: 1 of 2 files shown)


src/lenskit/knn/ease.py
```diff
@@ -0,0 +1,147 @@
...
+        log.debug("adding regularization term")
+        di = np.diag_indices(n_ok)
+        cooc[di] += self.config.regularization
+
+        log.debug("inverting Gram-matrix")
+        timer = Stopwatch()
+        mat = spla.inv(cooc, assume_a="pos")
+        log.info("inverted co-occurrance matrix in %s", timer)
+
+        mat /= np.diag(mat)
```
samiravaez: Aren't we missing a minus sign here?

src/lenskit/knn/ease.py
```diff
@@ -0,0 +1,147 @@
...
+        t_inos = items.numbers(vocabulary=self.items, missing="negative")
+        t_ok = t_inos >= 0
+        t_good = t_inos[t_ok]
+
+        N = len(self.items)
+        q_vec = np.zeros(N, dtype=np.float32)
+        q_vec[q_good] = 1.0
+
+        _log.debug("multiplying matrix for %d items", np.sum(q_ok))
+        scores = self.weights @ q_vec
```
samiravaez: Does this match the paper's s=xB scoring formula? Since we compute Bx here, I was wondering if the orientation is correct.

src/lenskit/knn/ease.py
mdekstrand: Good catch, yes!

src/lenskit/knn/ease.py
mdekstrand: Good catch, but since B is symmetric, it doesn't matter.

samiravaez: If P is symmetric, dividing it by its diagonal as in B = P / (-np.diag(P)) in the paper should break symmetry. I ran a quick test and it wasn’t symmetric. Am I missing something?

mdekstrand: You're probably right — I'll check again tomorrow.

mdekstrand: Yep, you're right. Fixed.

src/lenskit/training.py
```diff
@@ -102,6 +110,20 @@ def configured_device(self, *, gpu_default: bool = False) -> str:
...
+    def envvar(self, name: str, default: str) -> str: ...
+    @overload
+    def envvar(self, name: str, default: str | None = None) -> str | None: ...
+    def envvar(self, name: str, default: str | None = None) -> str | None:
+        """
+        Fetch a training environment variable.  Variables are first looked up in
+        :attr:`environment`, then in :attr:`os.environ`.
+        """
+        if name in self.environment:
+            return name
```
samiravaez, line 123: I think we should return the value here as self.environment[name], right?

mdekstrand: yep.

mdekstrand: Fix in flight in #1018.

PR #1032, opened 2026-03-05
Add Graded RBP metric class

### LINKED ISSUE #1031: Add graded rank-biased precision (GRBP)
Extend the existing `RBP` metric to handle _graded relevance judgements_ (non-binary `r_i ∈ [0, 1])` as described in [Rank-Biased Precision for Measurement of Retrieval Effectiveness ](https://dl.acm.org/doi/epdf/10.1145/1416950.1416952).  Graded RBP reflects fractional utility rather than binary success, but uses the same weighting and persistence model. 

Proposed Implementation
- Add a new class `GradedRBP`, inheriting from `RBP`
- Reads a relevance field (default: `grade`) from the test `ItemList`
- Scale grades to a unit range `[0, 1]`
- Default grade value for unseen items = `0.25`
- If `grade` field is not found, defaults to binary RBP

Questions

Should we assume the `grade` field is already scaled to [0, 1]? If not, we can apply scaling. So, should scaling be optional here? Should we divide by the max grade value for scaling or is there a different method?

sushobhan2024:
This PR is regarding the issue https://github.com/lenskit/lkpy/issues/1031

- The main change is around getting an array `relevance` instead of binary array. 
- `RBP` is now configurable with optional `grade_field`. The grades are assumed to be [0, 1] scaled. 
- Added a helper method `graded_rank_biased_precision` to compute scores using relevance and weights.
- Added tests in existing tests for `RBP`


src/lenskit/metrics/ranking/_rbp.py
```diff
@@ -158,4 +187,17 @@ def measure_list(self, recs: ItemList, test: ItemList) -> float:
...
+            good = recs.isin(test)
+            return rank_biased_precision(good, weights, normalization)
+
+        # Graded relevance
+        if self.grade_field not in test._fields:
+            raise ValueError(f"Grade field '{self.grade_field}' not found in test ItemList")
+
+        grades = test.field(self.grade_field)
+        grade_map = dict(zip(test.ids(), grades))
+        relevance = np.array([grade_map.get(item, self.unknown_grade) for item in recs.ids()])
```
samiravaez, line 201: Missing test items get unknown_grade with a default of 0.25, and the tests also check for that same default. Was that intentional? It doesn't seem righ to me

sushobhan2024: Good point. It probably makes sense to split this into two tests: one that verifies the default behavior, and another that verifies the parameter is actually being used.

samiravaez:
Also I'm not sure if we should have a normalization option for RBP, but if it's only for experimental purposes, I think it can exceed 1 with the current default for missing test items.
To get a lower bound score I think we should consider missing items' scores as zero. If we want an upper bound we could add the residual  p^k instead of assigning a fixed value.

src/lenskit/metrics/ranking/_rbp.py
```diff
@@ -109,6 +135,8 @@ def __init__(
         patience: float = 0.85,
         normalize: bool = False,
         weight_field: str | None = None,
+        grade_field: str | None = None,
+        unknown_grade: float = 0.25,
```
mdekstrand, line 139: `unknown_grade` should default to 0, 0.25 is specific to our experiment.

src/lenskit/metrics/ranking/_rbp.py
```diff
@@ -63,6 +84,9 @@ class RBP(ListMetric, RankingMetricBase):
     in the paper; however, RBP with high patience should be no worse than nDCG
     (and perhaps even better) in this regard.
 
+    This metric class supports relevance grades :math:`r_{ui} \\in \\[0, 1\\]`
+    via an optional ``grade_field``.
```
mdekstrand, line 88: need to add `grade_field` to the `Args:` documentation.

src/lenskit/metrics/ranking/_rbp.py
```diff
@@ -117,13 +145,16 @@ def __init__(
...
 
     @property
     def label(self):
+        base = "RBP" if self.grade_field is None else "GradedRBP"
         if self.n is not None:
-            return f"RBP@{self.n}"
+            return f"{base}@{self.n}"
         else:
-            return "RBP"
+            return base
```
mdekstrand, line 157: Good.

src/lenskit/metrics/ranking/_rbp.py
```diff
@@ -158,4 +187,17 @@ def measure_list(self, recs: ItemList, test: ItemList) -> float:
...
+        # Binary relevance
+        if self.grade_field is None:
+            good = recs.isin(test)
+            return rank_biased_precision(good, weights, normalization)
+
+        # Graded relevance
+        if self.grade_field not in test._fields:
+            raise ValueError(f"Grade field '{self.grade_field}' not found in test ItemList")
+
+        grades = test.field(self.grade_field)
```
mdekstrand, line 199: We usually should not poke around inside other classes' private data — `_fields` is not a public member of item list. It might be a good idea to add a public method to access the list of fields if we don't have one yet.

`field` will return `None` if the field is not found, so just call it and check the result for `None`.

mdekstrand (changes requested):
A couple things to fix, mostly looks good.

To @samiravaez's point on normalization: I also think removing NDCG-style normalization is probably a good idea — I'm not sure it's ever really the right thing to do. However, that should perhaps be a separate PR.

We also need to add to the release notes.

PR #1033, opened 2026-03-05
Add LightGCN support to FlexMF

### LINKED ISSUE #1019: Add LightGCN to FlexMF
It turns out that it would actually be pretty easy to add LightGCN as an option to FlexMF, rather than using Torch-Geometric's LightGCN. We should be able to implement it with significantly lower memory requirements.

This would supersede #848.

mdekstrand:
This updates FlexMF to have LightGCN support (for implicit feedback) that is activated with a new `convolution_layers` configuration setting. Zero layers gets classical matrix factorization.

Closes #1019.

### DIFF
--- a/docs/releases/2026.rst
+++ b/docs/releases/2026.rst
@@ -42,6 +42,7 @@ New Features
   Python 3.14t on Linux and macOS (:issue:`916`, :pr:`1022`).
 - Added support for parallel batch inference using thread pools on free-threaded
   Python (:issue:`921`, :pr:`1025`).
+- Added LightGCN support to FlexMF (:issue:`1019`, :pr:`1033`).
 
 Performance Changes
 -------------------


src/lenskit/flexmf/_model.py
```diff
@@ -129,15 +171,24 @@ def forward(self, user: Tensor, item: Tensor, *, return_norm: bool = False):
...
+        else:
+            uvec = uemb = self.u_embed(user)
 
         # compute the inner score
         ips = vecdot(uvec, ivec)
         score = ub + ib + ips
 
         if return_norm:
-            l2 = torch.square(ub) + torch.square(ib) + norm(uvec, dim=-1) + norm(ivec, dim=-1)
+            l2 = torch.square(ub) + torch.square(ib) + norm(uemb, dim=-1) + norm(iemb, dim=-1)
```
samiravaez, line 193: I think the uemb and iemb are the propagated embeddings, right? Shouldn't we only regularize intital embedings? They are the only trainable parameters

mdekstrand: Yes, only the trainable parameters should be included. This code has been through a couple iterations now but in the current version, `uemb` should be just the initial embeddings (`uvec` is the final propagated embedding).

src/lenskit/flexmf/_model.py
```diff
@@ -129,15 +171,26 @@ def forward(self, user: Tensor, item: Tensor, *, return_norm: bool = False):
...
+            ivec = self.i_layers[:, item, :].sum(0) * self._layer_scale
+            if return_norm:
+                iemb = self.i_layers[0, item, :]
+        else:
+            ivec = iemb = self.i_embed(item)
+
+        if self.u_layers is not None:
+            uvec = self.u_layers[:, user, :].sum(0) * self._layer_scale
+            if return_norm:
+                uemb = self.u_layers[0, item, :]
```
samiravaez: I think this should be indexed with user

src/lenskit/flexmf/_model.py
samiravaez: Also, I don't know if this will make a diference in practice but the paper uses the squared norm

mdekstrand: It might make a small difference, not a very large one though. However, we also rarely use this code path, and I might even remove after some more testing — my experiments so far indicate that AdamW weight decay works better than explicit regularization.

PR #1035, opened 2026-03-13
Add presets to FlexMF

mdekstrand:
This adds “presets” to the FlexMF model, and updates the documentation to describe them.


src/lenskit/flexmf/_implicit.py
```diff
@@ -28,6 +28,22 @@
...
+        "loss": "warp",
+        "negative_strategy": "misranked",
+        "user_bias": False,
+        "item_bias": False,
+    },
+    "lightgcn": {
+        "loss": "pairwise",
+        "user_bias": False,
+        "item_bias": False,
+        "convolution_layers": 2,
```
samiravaez: I think they set the # of layers to 3 for satisfactory performance

PR #1065, opened 2026-04-05
Filter negative gains in NDCG calculation

samiravaez:
Thi PR resolves #1057 by clipping negative gains to zero in both the computed DCG and the ideal DCG.


src/lenskit/metrics/ranking/_dcg.py
```diff
@@ -105,13 +105,17 @@ def measure_list(self, recs: ItemList, test: ItemList) -> float:
...
+            gains = gains.clip(lower=0)
             if self.n:
                 gains = gains.nlargest(n=self.n)
             else:
                 gains = gains.sort_values(ascending=False)
             iweight = self.weight.weight(np.arange(1, len(gains) + 1))
             ideal = np.dot(gains.values, iweight).item()  # type: ignore
 
+            if ideal == 0:
+                return 0.0
```
mdekstrand, line 122: yeah this is probably a good idea.

mdekstrand (approved):

mdekstrand:
Can we add a note on clipping the gains to the documentation?

PR #1093, opened 2026-05-09
feat(flexmf): add Neural Collaborative Filtering (NCF) model

danhdanhtuan0308:


src/lenskit/flexmf/_ncf.py
```diff
@@ -0,0 +1,166 @@
...
+class FlexMFNCFConfig(FlexMFConfigBase):
+    """
+    Configuration for NCF (Neural Collaborative Filtering) Scorer.
+
+    Stability:
+        Experimental
+    """
+
+    gmf_embedding_size: PositiveInt = 8
+    mlp_embedding_size: PositiveInt = 8
```
mdekstrand: In the current design, this class will inherit `embedding_size` from `FlexMFConfigBase`, but then not use it in favor of these two embedding sizes. Is there one of the embedding sizes that it would make sense to set to `embedding_size`, and then just have a different name for a second one?

src/lenskit/flexmf/_ncf.py
```diff
@@ -0,0 +1,166 @@
...
+    item_bias: bool = False
+    convolution_layers: int = 0
+
+    def selected_negative_strategy(self) -> NegativeStrategy:
+        if self.negative_strategy is not None:
+            return self.negative_strategy
+        elif self.loss == "warp":
+            return "misranked"
+        else:
+            return "uniform"
```
mdekstrand: This is a lot of copied code from the implicit model. Is there a reasonable way to enable more reuse?

(Also see the top-level review comment, though — that might be a cleaner solution to a whole set of problems.)

src/lenskit/flexmf/_ncf.py
```diff
@@ -0,0 +1,166 @@
...
+        # Ensure MLP inputs are broadcasted to the same shape before concatenating
+        if u_mlp.shape != i_mlp.shape:
+            u_mlp = u_mlp.expand(i_mlp.shape[:-1] + (u_mlp.shape[-1],))
+            i_mlp = i_mlp.expand(u_mlp.shape[:-1] + (i_mlp.shape[-1],))
+
+        gmf_out = u_gmf * i_gmf
+        mlp_out = self.mlp(torch.cat([u_mlp, i_mlp], dim=-1))
+
+        out = torch.cat([gmf_out, mlp_out], dim=-1)
+        score = self.prediction(out).squeeze(-1)
```
mdekstrand, line 143: What does this do?

src/lenskit/flexmf/_ncf.py
```diff
@@ -0,0 +1,166 @@
...
+    def selected_negative_strategy(self) -> NegativeStrategy:
+        if self.negative_strategy is not None:
+            return self.negative_strategy
+        elif self.loss == "warp":
+            return "misranked"
+        else:
+            return "uniform"
+
+
+class FlexMFNCFModel(FlexMFModel):
```
mdekstrand: Similar comment on inheritance as above - we're inheriting `FlexMFModel`, but then the attributes of this class are not a superset of the attributes of the base class. As a general rule, when we are inheriting a class, the derived class should extend the base class, not silently remove parts of it.

mdekstrand (commented):
I've got some comments on inheritance and a question about some unfamiliar PyTorch semantics, but there's a higher-level question: would it make the most sense to make this a separate class, or to make it a set of configuration options on the implicit scorer?

Or, alternatively, to make it more fully extend the implicit scorer, and reuse the implicit scorer's model as-is, combining it with a separate model/module implementing the MLP tower, and a third that combines the first two?

src/lenskit/flexmf/_ncf.py
danhdanhtuan0308: I am thinking adding mlp_layers and embedding_size to FlexMFimplicitConfig and extend FlexMFModel to to optional include the MLP tower which will make one scorer class handles both MF and NCF

src/lenskit/flexmf/_ncf.py
danhdanhtuan0308: It concatenates the GMF output and MLP outputs to produce a single score per user-item pair.

src/lenskit/flexmf/_ncf.py
danhdanhtuan0308: Sure I will change this

src/lenskit/flexmf/_ncf.py
```diff
@@ -0,0 +1,185 @@
...
+        super().__init__()
+
+        # GMF component: a standard MF model providing u_embed / i_embed.
+        self.gmf_model = FlexMFModel(
+            e_size=gmf_e_size,
+            n_users=n_users,
+            n_items=n_items,
+            rng=rng,
+            user_bias=False,
+            item_bias=False,
```
mdekstrand, line 81: I understand changing the defaults (although that may be a little confusing to properly document), but is there a reason we shouldn't allow user and item biases?

PR #1096, opened 2026-05-16
fix graded NDCG crashes for invalid gain values

### LINKED ISSUE #1095: Handle missing or invalid gain data in graded NDCG
Graded NDCG crashes when a test list is empty or missing the gain field.

samiravaez:
This PR resolves #1095. Graded NDCG now returns `NaN` instead of raising errors when gain values are missing or null. Null values are considered unknown and excluded from the computation.


src/lenskit/metrics/ranking/_dcg.py
```diff
@@ -102,21 +104,28 @@ def measure_list(self, recs: ItemList, test: ItemList) -> float:
...
-            return float("nan")
+            return np.nan
 
         if self.gain:
             realized = _graded_dcg(recs, test, self.gain, self.weight)
 
             gains = test.field(self.gain, "pandas", index="ids")
             if gains is None:
-                raise KeyError(f"test items have no field {self.gain}")
+                warnings.warn(f"test items have no field {self.gain}")
```
mdekstrand: This should be a warning of type `DataWarning`. Also, do we want to consider still throwing the error in the case when `test` is not empty? I.e., if there is test data, but it has no field, it's an error; if the test data is fully empty, it's just a warning.

mdekstrand (approved):

PR #1162, opened 2026-07-27
Explicitly handle time zones in time-filtering builders

mdekstrand:
This updates the `DatasetBuilder` logic to explicitly handle time zones when filtering by timestamp, to ensure handling is unambiguous.


mdekstrand:
This was actually merged, but a small error mis-tagged it.

PR #1180, opened 2026-08-04
Run dataset tests on Woodpecker with full data access

mdekstrand:
This adds a Woodpecker CI job with locally-cached access to datasets for more thorough testing.


mdekstrand:
Cancelled in favor of #1181.

PR #1182, opened 2026-08-06
Add IPS-weighted RBP (#1155)

### LINKED ISSUE #1155: Add IPS-weighted RBP
We have an implementation of RBP with primitive IPS weighting in an experiment, and should add it to LensKit.

vvineis:
Adds IPS-weighted RBP (#1155), following Yang et al., "Unbiased Offline Recommender Evaluation for Missing-Not-At-Random Implicit Feedback" (RecSys 2018).

The metric is `IPSRBP`. It uses the paper's self-normalized estimator (SNIPS, their eqn. 9) by default; `self_normalized=False` drops the denominator and gives the plain IPS numerator. Propensities come from a `PropensityModel` passed to the constructor: `PopularityPropensity` implements their eqn. 15, `UniformPropensity` is the missing-at-random baseline, and `FieldPropensity` reads them off a field if you estimated them yourself. There's also `estimate_power_law_gamma`, which fits the exponent from a set of recommendations the way they do in section 4.3.

I put the propensities in a separate object rather than as arguments on the metric, partly to follow `MeanPopRank` and `ListGini` taking a `Dataset`, and partly because it mirrors `RankWeight`. IPS-corrected Recall or DCG should then just be a matter of reusing it.

A few choices:
- `gamma` is the paper's exponent, so the `(gamma+1)/2` of eqn. 15 happens inside the propensity model and values from their Table 2 can be used as-is. It has no default, since it depends on the dataset and on whatever produced the log. 
- The paper only pins propensities up to a constant. I normalize by the most popular item, so P lands in (0, 1] and weights are at least 1. It makes no difference under SNIPS, but the unnormalized form needs a fixed scale.

## Tests

Two identities tie the metric to the existing one, both checked with Hypothesis over random lists, cutoffs and patience values:
- uniform propensities with `self_normalized=False` is exactly `RBP`
- uniform propensities with SNIPS is exactly the paper's average-over-all evaluator (eqn. 6), which for RBP works out to `RBP / |S*_u|`

I also checked both per user on ml-latest-small and they hold to machine precision.

There's also a small simulated MNAR log where the true reward of eqn. 1 is known: the uncorrected evaluator overstates a popularity-chasing recommender and the correction pulls it back, while on a missing-at-random log the correction leaves the score alone.

## One thing I didn't do

The docstrings cite `yangUnbiasedOfflineRecommender2018`, which isn't in `docs/lenskit.bib` yet. 

@inproceedings{yang2018unbiased,
  title={Unbiased offline recommender evaluation for missing-not-at-random implicit feedback},
  author={Yang, Longqi and Cui, Yin and Xuan, Yuan and Wang, Chenyang and Belongie, Serge and Estrin, Deborah},
  booktitle={Proceedings of the 12th ACM conference on recommender systems},
  pages={279--287},
  year={2018}
}


mdekstrand:
Will provide a more detailed review tomorrow or Monday, but I just cut the 2026.3 release, so this will be targeted for 2026.4.0.

docs/guide/evaluation/rankings.rst
```diff
@@ -75,9 +75,37 @@ or graded relevance scores).
...
+
+    propensity = PopularityPropensity(split.train, gamma=1.87)
+    mc.add_metric(RBP(n=10))
+    mc.add_metric(IPSRBP(n=10, propensity=propensity))
+
+Estimate propensities from the training data, not from the test data being
+measured: the unbiasedness argument holds them fixed with respect to the
+observations they weight.
+
+.. versionadded:: 2026.3
```
mdekstrand, line 107: bump to 2026.4

docs/releases/2026.rst
```diff
@@ -46,6 +46,9 @@ fixed issues and merged PRs is on `milestone 2026.3`_.
 - Certain per-request warnings in batch inference are now aggregated (:issue:`1113`, :pr:`1168`).
 - Added :class:`lenskit.metrics.UniqueItemCount` (:issue:`1005`, :pr:`1169`).
 - Added support for ``interaction="default"`` to :meth:`~lenskit.data.DatasetBuilder.add_relationship_class` (:pr:`1174`).
+- Added :class:`lenskit.metrics.IPSRBP`, an IPS-weighted rank-biased precision
+  for missing-not-at-random implicit feedback, along with the propensity models
+  it uses (:issue:`1155`).
```
mdekstrand, line 51: Move to new section for 2026.4.

src/lenskit/metrics/ranking/_propensity.py
```diff
@@ -0,0 +1,289 @@
...
+
+    Stability:
+        Caller
+    """
+    if not 0 <= exclude_quantile < 1:
+        raise ValueError("exclude_quantile must be in [0, 1)")
+
+    counts = _item_counts(data, count)
+    counts = counts[counts > 0]
+    shown = [items[:n].ids() for _, items in recs if len(items)]
```
mdekstrand, line 253: this can be `for items in recs.lists()`.

mdekstrand (approved):
Thanks! Design looks good to me.

The thing I currently find unclear is `gamma`, and how to choose it, especially when we don't know the logging policy. Do we always need to use `estimate_power_law_gamma`, or is there a reasonable way to pick some default value? If we always need to use `estimate_power_law_gamma`, how should we generate the recommendations to do so?

Issue #1142, opened 2026-07-09
Add AMBAR dataset importer and CLI support


sushobhan2024:
Add support for importing the AMBAR dataset into LensKit.

This includes:

- [x] Add a `load_ambar()` dataset loader that reads the AMBAR dataset and returns a `Dataset`.
- [x] Add CLI support so AMBAR can be imported using the LensKit command-line tools.
- [x] Update the CLI so imported dataset can be described with the existing `describe` command.

mdekstrand:
Implemented in #1187.

PR #1187, opened 2026-08-18
Add AMBAR music dataset to Lenskit

sushobhan2024:
issue #1142

- Added `amber.py` to load data from its path
- Updated the `describe.py` so imported dataset can be described with the existing describe command


mdekstrand (changes requested):
A few things this needs off the bat:

- tests (tagged with `@mark.realdata`, see the Amazon and Steam load tests for examples). The tests should probably look at it in `data/ambar`.
- adding to `data/convert.py`.
- adding to the data conversion CLI test script.

mdekstrand:
Now that the tests are there, they are failing: https://jynx.mde.one/repos/1/pipeline/54/3

The test environment has a copy of the `AMBAR` folder from the data set repo at `data/ambar`.

sushobhan2024:
The error is due to duplicate `user_ids`. There are 8 duplicate user_ids.
5 have conflicts (contradictory country/continent, one even contradictory gender). The other 3 are repeated records.
I am considering removing those 5 records, and keeping first of the 3 repeated records.

Also, there are 8 different `user_ids` in `ratings_info.csv` that have no row at all in `users_info.csv`. Removing these as well, and passing `missing="error"` instead of `missing="insert"` in `add_interactions()`.

> conflicting user_ids (from users_info duplicates): 5
{506586384, 39168978, 128783640, 390604169, 38606714}

> truly missing user_ids (in ratings, absent from users_info): 8
{271316318, 306287460, 163187558, 839347986, 871284436, 955505332, 183305018, 387251006}

> overlap between the two groups: 0

> ratings from conflicting users: 1018
> ratings from truly missing users: 747
> total distinct user_ids being excluded: 13

1,765 of 3,311,462 total ratings (0.053%) from these 13 users are now excluded.

Let me know if excluding is the right approach here.

mdekstrand:
Probably better to keep the ratings with missing users, those users just won't have attributes. On the conflicts, dropping is probably fine, we could also keep them and just pick the first or last record (if the records have timestamps, we should pick the last one by timestamp).

mdekstrand:
@mdekstrand just FYI, for security reasons I have to manually approve the woodpecker tests every time they run. They can't be re-triggered from within a push.

sushobhan2024:
This dataset does not have timestamps.

mdekstrand:
@sushobhan2024 the tests are running again now.

mdekstrand (approved):

PR #1201, opened 2026-09-01
add query-time pooling for flexMF

### LINKED ISSUE #934: Support pooling-based implicit FlexMF
The first and simplest mechanism to use the query-time user history to score and recommend in `FlexMF` (#1198) is to *mean-pool* the embeddings of the query items.

That is, we get the query items from the `RecQuery`, and take the average of the item embeddings to use as the user embedding.

This should be configurable in a similar way as ALS's use of item embeddings: we should configure whether to prefer the pre-trained embedding or the pooled embedding when a known user ID arrives with history.

The BPR and WARP presets, and possibly also LightGCN, should default to preferring the pre-trained.

> [!NOTE]
>
> Explicit feedback requires more subtlety than just pooling, so this issue only covers implicit feedback.

samiravaez:
This PR resolves #934 by adding responsive recommendations to FlexMF.
FlexMF can now build a temporary user representation at scoring time by mean-pooling the embeddings of the items in `RecQuery.query_items`. This allows recommendations to be generated from query context even when there is no trained user embedding available.
Unknown query items are ignored, and if pooling cannot produce a usable vector, the model falls back to the trained user embedding when allowed. Explicit model scoring also preserves the global bias when using the pooled-user path

### DIFF
--- a/src/lenskit/flexmf/_base.py
+++ b/src/lenskit/flexmf/_base.py
@@ -103,6 +103,7 @@ class FlexMFConfigBase(EmbeddingSizeMixin, BaseModel):
     query provides items.
     """
 
+
 class FlexMFScorerBase(UsesTrainer, Component):
     """
     Base class for the FlexMF scorers, providing common Torch support.
@@ -135,7 +136,7 @@ class FlexMFScorerBase(UsesTrainer, Component):
         u_row = None
         if query.user_id is not None:
             u_row = self.users.number(query.user_id, missing=None)
-        
+
         # make sure it's on the right device
         device = self.model.device
         u_tensor = None
@@ -149,15 +150,13 @@ class FlexMFScorerBase(UsesTrainer, Component):
         pool_query = (
             query_items is not None
             and len(query_items) > 0
-            and (u_row is None or self.config.user_embeddings != "prefer"))
+            and (u_row is None or self.config.user_embeddings != "prefer")
+        )
 
         pooled_user = None
         if pool_query:
             # resolve query items against the model's item vocabulary
-            q_cols = query_items.numbers(
-                vocabulary=self.items,
-                missing="negative",
-                format="torch")
+            q_cols = query_items.numbers(vocabulary=self.items, missing="negative", format="torch")
             q_cols = q_cols.to(device, non_blocking=True)
 
             # ignore query items that were not known during model training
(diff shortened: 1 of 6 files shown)


src/lenskit/flexmf/_base.py
```diff
@@ -123,20 +132,45 @@ def __call__(self, query: QueryInput, items: ItemList) -> ItemList:
...
+
+        # decide whether query items should provide the user embedding
+        query_items = query.query_items
+        pool_query = (
+            query_items is not None
+            and len(query_items) > 0
+            and (u_row is None or self.config.user_embeddings != "prefer")
+        )
+
+        pooled_user = None
```
mdekstrand: We can just reuse the `u_tensor`, can't we? That will also enable making the `if` on line 172 simpler.

src/lenskit/flexmf/_base.py
```diff
@@ -123,20 +132,45 @@ def __call__(self, query: QueryInput, items: ItemList) -> ItemList:
...
+            q_cols = q_cols.masked_select(q_cols.ge(0))
+
+            if len(q_cols) > 0:
+                # mean-pool the item embeddings to obtain a query-time
+                # representation of the user
+                q_vectors = self.model.i_embed(q_cols)
+                pooled_user = q_vectors.mean(dim=0)
+
+        # if pooling was not possible, fall back to the trained user embedding
+        if pooled_user is None and (u_tensor is None or not self.config.user_embeddings):
```
mdekstrand: If we reuse `u_tensor` above, then this can just be `if u_tensor is None`. I.e., the logic becomes "there are several ways to get `u_tensor`, if we've tried all of them and not gotten a `u_tensor`, then we can't score".

It is possible that this will also fix the test failure — I'm not sure if all of the `if` logic in this method is correct yet, partly because it seems more complicated than it needs to be.

src/lenskit/flexmf/_base.py
```diff
@@ -162,3 +201,9 @@ def score_items(self, users: torch.Tensor, items: torch.Tensor) -> torch.Tensor:
         users and items.
         """
         return self.model(users, items)
+
+    def score_user_embedding(self, user: torch.Tensor, items: torch.Tensor) -> torch.Tensor:
+        """
+        Score items against a user embedding.
+        """
+        return self.model.score_user_vector(user, items)
```
mdekstrand: I'm not sure why we need a new function here - isn't the final result of pooling just a user embedding, like any other? That was the intention with the issue. (Whether it is a _good_ user embedding is another question.)

samiravaez: what the user bias should be for the pooled embedding if I use the same scoring function?

mdekstrand: User bias is an important and subtle question here, and it might actually be different for explicit and implicit feedback models.

For implicit-feedback models, using 0 as the fresh user bias is probably fine. It might miscalibrate probabilities, but won't affect ranking.

For explicit-feedback models, I think we have 3 choices:

- use 0
- use saved user bias, even if we aren't using saved user embedding
- estimate user bias from rating values before computing the pooled embedding

mdekstrand (approved):

PR #1210, opened 2026-09-07
Convert numeric interaction timestamps

### LINKED ISSUE #1183: Convert timestamp columns to timestamps in DatasetBuilder
`DatasetBuilder.add_interactions` (and `from_interactions_df`) should have better handling for timestamp columns, since those are recognized specially by LensKit:

- [ ] Convert timestamp columns to Timestamp types.
- [ ] Provide an argument to control timestamp units.
- [ ] Guess timestamp units from numeric timestamps when units are not specified, aiming to put timestamps in the 1900-2000s.

appleweiping:
Fixes #1183.

This converts numeric interaction timestamps to Arrow timestamp types in both `DatasetBuilder.add_interactions` and `from_interactions_df`.

- adds an explicit `timestamp_unit` option
- infers integer values from seconds through nanoseconds when at least 95 percent fall between 1900 and 2099
- treats floating point values as seconds by default, preserves millisecond precision, and converts NaN to null
- preserves existing timestamp columns and promotes compatible units across batches
- keeps timestamp attributes usable in SciPy sparse matrices as integer seconds

Validation:

- data test suite: 438 passed, 40 skipped, 3 expected failures
- focused timestamp and builder tests: 64 passed
- Ruff lint and format checks passed

Developed with OpenAI Codex assistance. I reviewed and tested the change and remain responsible for it.

### DIFF
--- a/src/lenskit/data/_builder.py
+++ b/src/lenskit/data/_builder.py
@@ -51,7 +51,6 @@ _log = get_logger(__name__)
 
 type TableInput = pd.DataFrame | pa.Table | dict[str, NDArray[Any]]
 type RelationshipEntities = Sequence[str] | Mapping[str, str | None]
-type TimestampUnit = Literal["s", "ms", "us", "ns"]
 
 type DuplicateAction = Literal["update", "error", "overwrite"]
 """
@@ -392,6 +391,7 @@ class DatasetBuilder:
         interaction: bool | Literal["default"] = False,
         _warning_parent: int = 0,
         remove_repeats: bool | Literal["exact"] = False,
+        timestamp_unit: Literal["s", "ms", "us", "ns"] | None = None,
     ) -> None:
         """
         Add relationship records to the data set.
@@ -424,6 +424,10 @@ class DatasetBuilder:
             remove_repeats:
                 If ``True``, repeated interactions will be removed. If ``"exact"``,
                 duplicated interactions will be removed.
+            timestamp_unit:
+                The unit of numeric values in the ``timestamp`` column. If omitted,
+                integer units are inferred and floating-point values are assumed to
+                be seconds. Numeric timestamps are converted to Arrow timestamps.
         """
         if isinstance(data, pd.DataFrame):
             table = pa.Table.from_pandas(data, preserve_index=False)
@@ -431,6 +435,8 @@ class DatasetBuilder:
             table = pa.table(data)  # type: ignore
         else:
             table = data
+
+        table = _convert_relationship_timestamps(table, timestamp_unit)
 
         log = self._log.bind(class_name=cls, count=table.num_rows)
 
@@ -571,7 +577,7 @@ class DatasetBuilder:
         allow_repeats: bool = True,
         default: bool = False,
         remove_repeats: bool | Literal["exact"] = False,
-        timestamp_unit: TimestampUnit | None = None,
+        timestamp_unit: Literal["s", "ms", "us", "ns"] | None = None,
     ) -> None:
         """
         Add a interaction records to the data set.
@@ -614,7 +620,6 @@ class DatasetBuilder:
                 integer units are inferred and floating-point values are assumed to
                 be seconds. Numeric timestamps are converted to Arrow timestamps.
         """
-        data = _convert_interaction_timestamps(data, timestamp_unit)
         self.add_relationships(
             cls,
             data,
@@ -624,6 +629,7 @@ class DatasetBuilder:
             interaction="default" if default else True,
             _warning_parent=1,
             remove_repeats=remove_repeats,
+            timestamp_unit=timestamp_unit,
         )
 
     def filter_interactions(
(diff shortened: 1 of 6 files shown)


appleweiping:
The dev container check failed in tests/tuning/test_optuna_point.py::test_float_points with a Hypothesis FlakyFailure. A generated log scale float fell about 0.0011 below its lower bound, then the failure did not reproduce on Hypothesis retry. This test is outside the timestamp code changed here. I do not have permission to rerun upstream Actions. Could a maintainer rerun the failed job when convenient?

tests/data/test_builder_interactions.py
```diff
@@ -135,6 +135,124 @@ def test_add_interactions_table():
     assert np.all(mat.rowptrs == [0, 2, 3, 6])
 
 
+@mark.parametrize(
+    ("unit", "value"),
+    [
+        ("s", 1_700_000_000),
```
mdekstrand: There are quite a few magic numbers in the new tests in this file — what is the purpose of selecting these values specifically? Please add code comments describing what these values are chosen to accomplish.

src/lenskit/data/_builder.py
```diff
@@ -570,6 +571,7 @@ def add_interactions(
         allow_repeats: bool = True,
         default: bool = False,
         remove_repeats: bool | Literal["exact"] = False,
+        timestamp_unit: TimestampUnit | None = None,
```
mdekstrand: Let's put `Literal[...]` directly in the function signatures instead of defining a separate type — that way the acceptable values will be directly in the method documentation.

src/lenskit/data/_builder.py
```diff
@@ -607,7 +609,12 @@ def add_interactions(
             remove_repeats:
                 If ``True``, repeated interactions will be removed. If ``"exact"``,
                 duplicated interactions will be removed.
+            timestamp_unit:
+                The unit of numeric values in the ``timestamp`` column. If omitted,
+                integer units are inferred and floating-point values are assumed to
+                be seconds. Numeric timestamps are converted to Arrow timestamps.
         """
+        data = _convert_interaction_timestamps(data, timestamp_unit)
```
mdekstrand: We need to add the `timestamp_unit` option to `add_relationships` as well, and move the conversion logic into that method. It is valid to use `add_relationships` to add interactions, and the timestamp logic should be identical.

src/lenskit/data/_builder.py
```diff
@@ -1228,6 +1235,84 @@ def _empty_rel_table(types: list[str]) -> pa.Table:
...
 
+_TIMESTAMP_UNITS: tuple[TimestampUnit, ...] = ("s", "ms", "us", "ns")
+_TIMESTAMP_UNIT_FACTORS: dict[TimestampUnit, int] = {
+    "s": 1,
+    "ms": 1_000,
+    "us": 1_000_000,
+    "ns": 1_000_000_000,
+}
+_TIMESTAMP_MIN_SECONDS = -2_208_988_800
+_TIMESTAMP_MAX_SECONDS = 4_102_444_800
```
mdekstrand: For clarity, it would be better to define these by creating `datetime` objects from ISO date strings and extracting the UNIX timestamp, so that it is obvious what these values mean when they are defined.

mdekstrand (changes requested):
Thank you for submitting this PR! The dev container test is flaky right now for unclear reasons, so I'm not worried about that one failing if everything else is passing. I've also kicked off the data CI tests (those require manual approval).

The strategy looks sound. I've added comments on some places where we need additional documentation, and one logic change (`add_relationships` needs to support timestamp inference too).

appleweiping:
Thanks for the review. I moved timestamp conversion into add_relationships and added timestamp_unit there. add_interactions now forwards the option. I also inlined the accepted values in the public signatures, replaced the numeric bounds with ISO datetimes, and explained the test values. I added direct add_relationships coverage as well. The complete data test suite passed with 440 tests passed, 42 skipped, and 3 expected failures. Ruff checks passed.

appleweiping:
The new test jobs are failing before project setup because the mise action is attempting to download version 2026.9.3 and receives a 404. LensKit tests do not start. The local full data suite remains green.

appleweiping:
The previous rerun completed with all project and dataset tests passing. Codecov's project check remained red on a 0.01% overall coverage change even though patch coverage passed.

I added direct coverage for the remaining timestamp conversion branches: an explicit integer unit, an all-null timestamp column, a nonnumeric timestamp column, and an invalid unit. The updated interaction test file passes all 26 tests, and Ruff passes.

appleweiping:
All four review comments are addressed in the current head, including direct add_relationships coverage and the added edge cases. The full project and dataset test suites passed, and patch coverage is green. The remaining Codecov project result is a 0.01 percent repository wide change. Please re-review when convenient.

mdekstrand (approved):

